PR #775, opened 2026-04-21
fix(773): DataStorage CRD UPDATE Re-Sync + SOC2 Audit Compliance

### LINKED ISSUE #773: DataStorage does not re-sync RemediationWorkflow CRD updates to its internal catalog
## Summary

When a `RemediationWorkflow` CRD is updated via `kubectl apply` (e.g., changing `spec.description.whenNotToUse`), the Kubernetes API server reflects the change immediately, but DataStorage's internal PostgreSQL workflow catalog continues serving the **stale** text to callers of `list_workflows`. A DataStorage pod restart is required to force a re-sync.

## Steps to reproduce

1. Apply an updated `RemediationWorkflow` CRD:
   ```bash
   kubectl apply -f deploy/remediation-workflows/crashloop/crashloop.yaml
   # Output: remediationworkflow.kubernaut.ai/crashloop-rollback-v1 configured
   ```

2. Verify the CRD is updated in the cluster:
   ```bash
   kubectl get remediationworkflow crashloop-rollback-v1 -n kubernaut-system \
     -o jsonpath='{.spec.description.whenNotToUse}'
   # Output: <new text>
   ```

3. Trigger a scenario that causes KA to call `list_workflows` for `RollbackDeployment`.

4. Inspect the audit event for that `list_workflows` tool call:
   ```sql
   SELECT event_data::text FROM audit_events
   WHERE event_type='aiagent.llm.tool_call'
     AND event_data->>'tool_name' = 'list_workflows'
   ORDER BY event_timestamp DESC LIMIT 1;
   ```
   The `tool_result.workflows[].description.whenNotToUse` still contains the **old** text.

5. Restart DataStorage:
   ```bash
   kubectl rollout restart deployment/datastorage -n kubernaut-system
   ```

6. Re-run the scenario. The `list_workflows` response now contains the **new** text.

## Expected behavior

DataStorage's CRD controller/watch should detect `RemediationWorkflow` updates (not just creates/deletes) and re-sync the changed fields into its PostgreSQL workflow catalog without requiring a pod restart.

## Actual behavior

The reconciliation loop either:
- Does not trigger on CRD spec updates (only on create/delete), or
- Runs on a periodic interval that is too long, or
- Skips re-syncing description fields when the workflow already exists in the DB

## Impact

This caused the `concurrent-cross-namespace` scenario to fail validation. The `whenNotToUse` guidance for `crashloop-rollback-v1` was updated to steer the LLM toward `PatchConfiguration` when `risk_tolerance=high`, but the LLM never saw the new text. After restarting DataStorage, the updated text was served and the LLM correctly queried both `PatchConfiguration` and `RollbackDeployment` action types.

**Any workflow metadata change (description, labels, customLabels) is invisible to KA until DataStorage is restarted**, which is undocumented and error-prone.

## Evidence

DataStorage logs from the second run (after restart) confirm the LLM queried both action types:
```
17:29:39 Workflows listed by action type  action_type=PatchConfiguration  total_count=1
17:29:42 Workflows listed by action type  action_type=RollbackDeployment  total_count=3
```

In the first run (before restart), only `RollbackDeployment` was queried because the old `whenNotToUse` text did not mention `risk_tolerance` or `PatchConfiguration`.

## Context

- Discovered during `concurrent-cross-namespace` scenario validation in kubernaut-demo-scenarios
- Related: jordigilh/kubernaut-demo-scenarios#334 (LLM selects wrong workflow for high risk_tolerance)
- DataStorage source: `pkg/datastorage/server/workflow_handlers.go`
- Platform helper comment (line 466): "The DataStorage controller reconciles it into the workflow catalog."

jordigilh:
## Summary

Fixes #773 — RemediationWorkflow CRD UPDATE operations were not intercepted by the E2E webhook manifest (missing `UPDATE` in operations list), the `handleUpdate` handler used best-effort semantics violating SOC2 CC8.1, audit events conflated UPDATE with CREATE, and DS incorrectly superseded when receiving same (name, version) with different content instead of rejecting the request.

### Changes across 4 workstreams:

- **SOC2 Audit Compliance**: Add `remediationworkflow.admitted.update` event type to OpenAPI spec + regenerate ogen client. Successful UPDATEs emit distinct audit events (not CREATE).
- **Security Hardening**: `handleUpdate` now matches `handleCreate` strictness — all error paths return `admission.Denied` with `emitDeniedAudit`. Extracted shared `registerWorkflow()` method to DRY the identical flow.
- **DS Content Integrity**: When an active workflow with the same (name, version) but different content hash is submitted, DS returns 409 Conflict (`content-integrity-violation`) instead of silently superseding. Version bump is required to register new content.
- **E2E Manifest + Tests**: Add `UPDATE` to E2E `ValidatingWebhookConfiguration` operations (matching Helm chart). New E2E tests cover version bump propagation, same-version content rejection, and idempotent re-apply. Added Helm smoke test for webhook operations.

### Governance:
- BR-WORKFLOW-006 v1.1: Replace "UPDATE: Passthrough" with full re-registration semantics
- DD-WORKFLOW-012 v2.2: Document version-locked content immutability
- DD-WEBHOOK-001 v1.4: Update matrix from CREATE/DELETE to CREATE/UPDATE/DELETE

## Test plan

- [x] All 166 AuthWebhook unit tests pass (including 5 new UT-AW-773-* tests)
- [x] All 653 DataStorage unit tests pass (including updated UT-DS-INTEGRITY-* and UT-DS-548-008)
- [x] Full codebase `go build ./...` passes
- [x] `go vet` clean on all changed packages
- [x] Pre-commit anti-pattern hooks pass
- [ ] E2E tests (E2E-AW-773-001/002/003) — require Kind cluster with updated services
- [ ] Helm smoke test ST-WEBHOOK-OPS-001 — requires chart rendering environment

### DIFF
--- a/api/openapi/data-storage-v1.yaml
+++ b/api/openapi/data-storage-v1.yaml
@@ -2513,6 +2513,7 @@ components:
               'webhook.remediationapprovalrequest.decided': '#/components/schemas/RemediationApprovalAuditPayload'
               'webhook.remediationrequest.timeout_modified': '#/components/schemas/RemediationRequestWebhookAuditPayload'
               'remediationworkflow.admitted.create': '#/components/schemas/RemediationWorkflowWebhookAuditPayload'
+              'remediationworkflow.admitted.update': '#/components/schemas/RemediationWorkflowWebhookAuditPayload'
               'remediationworkflow.admitted.delete': '#/components/schemas/RemediationWorkflowWebhookAuditPayload'
               'remediationworkflow.admitted.denied': '#/components/schemas/RemediationWorkflowWebhookAuditPayload'
               'workflow.catalog.actions_listed': '#/components/schemas/WorkflowDiscoveryAuditPayload'
@@ -4278,12 +4279,13 @@ components:
       required: [event_type, workflow_name, action]
       description: |
         Audit payload for RemediationWorkflow CRD admission events (ADR-058).
-        Emitted by the authwebhook when a RemediationWorkflow CRD is created, deleted, or denied.
+        Emitted by the authwebhook when a RemediationWorkflow CRD is created, updated, deleted, or denied.
       properties:
         event_type:
           type: string
           enum:
             - 'remediationworkflow.admitted.create'
+            - 'remediationworkflow.admitted.update'
             - 'remediationworkflow.admitted.delete'
             - 'remediationworkflow.admitted.denied'
           description: Event type for discriminator (matches parent event_type)
@@ -4293,7 +4295,7 @@ components:
           example: "crashloop-rollback-v1"
         action:
           type: string
-          enum: [create, delete, denied]
+          enum: [create, update, delete, denied]
           description: Admission action performed
           example: "create"
         workflow_id:
--- a/docs/architecture/decisions/DD-WEBHOOK-001-crd-webhook-requirements-matrix.md
+++ b/docs/architecture/decisions/DD-WEBHOOK-001-crd-webhook-requirements-matrix.md
@@ -7,6 +7,7 @@
 **Scope**: All Kubernetes CRDs in Kubernaut requiring user authentication
 
 **Version History**:
+- **v1.4** (April 21, 2026): Issue #773 — RemediationWorkflow operations updated from CREATE/DELETE to CREATE/UPDATE/DELETE. UPDATE now triggers DS re-registration with content integrity enforcement (409 on same version + different content). Distinct `remediationworkflow.admitted.update` audit event type added (SOC2 CC8.1).
 - **v1.3** (March 4, 2026): Added RemediationWorkflow (CREATE/DELETE) as 4th webhook handler. Workflow registration now uses CRD + ValidatingWebhook (ADR-058), replacing REST-only approach. Corrects v1.1 note about workflow CRUD using HTTP middleware.
 - **v1.2** (January 6, 2026): **ARCHITECTURE UPDATE**: Single consolidated webhook deployment (`kubernaut-auth-webhook`) with multiple handlers. Updated implementation approach and timelines. Added references to comprehensive implementation and test plans.
 - **v1.1** (January 6, 2026): Added NotificationRequest (DELETE attribution). ~~Note: Workflow CRUD uses HTTP middleware, not CRD webhook~~ (Corrected in v1.3: RemediationWorkflow now uses CRD webhook)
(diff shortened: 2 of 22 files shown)


jordigilh:
## E2E-AW-773-001 Failure: Race condition in `updateCRDStatus` on UPDATE path

### Triage

The only failure is `E2E-AW-773-001: version bump UPDATE propagates to DS catalog` — timed out after 30s waiting for `Status.WorkflowID` to change after a version bump. The 18 pre-existing tests all pass, as do the other two new tests (E2E-AW-773-002, E2E-AW-773-003).

### Root Cause

There is a race between the async `updateCRDStatus` goroutine and the API server persisting the UPDATE to etcd. The webhook is a **validating admission webhook** — it executes *before* the object is committed.

**CREATE path (works):**
1. Webhook admits CREATE, starts `go updateCRDStatus(...)`, returns `Allowed`
2. `RetryGetCRD` is called — CRD **doesn't exist yet** in etcd, so it retries with 500ms backoff
3. By the time `RetryGetCRD` succeeds, the API server has committed the CREATE (correct `resourceVersion`)
4. `Status().Update()` succeeds ✅

**UPDATE path (fails):**
1. Webhook admits UPDATE (version bump), starts `go updateCRDStatus(...)`, returns `Allowed`
2. `RetryGetCRD` is called — CRD **already exists**, so it succeeds **immediately** on the first attempt, reading the **pre-update** version (stale `resourceVersion`)
3. API server then commits the spec change to etcd (bumps `resourceVersion`)
4. `Status().Update()` uses the stale `resourceVersion` → **409 Conflict** → silently logged and dropped (no retry)
5. `Status.WorkflowID` is never updated → test times out

The `RetryGetCRD` helper was designed for the CREATE race (CRD not in etcd yet) but doesn't account for the UPDATE race (CRD exists with stale `resourceVersion`). And `updateCRDStatus` has **no `RetryOnConflict`** wrapper — a single 409 kills the status update silently.

### Suggested Fix

Replace the single-shot GET + status write with a standard Kubernetes optimistic concurrency loop using `retry.RetryOnConflict`:

```go
import "k8s.io/client-go/util/retry"

func (h *RemediationWorkflowHandler) updateCRDStatus(namespace, name, registeredBy string, result *WorkflowRegistrationResult) {
	logger := ctrl.Log.WithName("rw-webhook").WithValues("operation", "status-update", "name", name, "namespace": namespace)

	if h.k8sClient == nil {
		logger.Info("k8sClient not configured — skipping status update")
		return
	}

	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
	defer cancel()

	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
		rw := &rwv1alpha1.RemediationWorkflow{}
		if err := h.k8sClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, rw); err != nil {
			return err
		}

		now := metav1.Now()
		rw.Status.WorkflowID = result.WorkflowID
		rw.Status.CatalogStatus = sharedtypes.CatalogStatus(result.Status)
		rw.Status.RegisteredBy = registeredBy
		rw.Status.RegisteredAt = &now
		rw.Status.PreviouslyExisted = result.PreviouslyExisted

		return h.k8sClient.Status().Update(ctx, rw)
	})
	if err != nil {
		logger.Error(err, "Failed to update CRD status after retries",
			"workflow_id", result.WorkflowID,
		)
		return
	}

	logger.Info("CRD status updated",
		"workflow_id", result.WorkflowID,
		"catalog_status", result.Status,
		"previously_existed", result.PreviouslyExisted,
	)
}
```

On a 409, `RetryOnConflict` re-invokes the closure which re-GETs the CRD (getting the post-UPDATE `resourceVersion`) and retries the status write. This is the standard K8s pattern for optimistic concurrency and fixes both the UPDATE race and any future race with the reconciler's finalizer mutations.

Note: `RetryGetCRD` can be removed from this method since the retry loop's `Get` already handles the not-found backoff implicitly (the `DefaultBackoff` starts at 10ms). If you want to preserve the CREATE-specific not-found retry behavior, you can keep `RetryGetCRD` as the initial GET before entering the conflict retry loop, but the simpler approach above covers both cases.

Issue #1012, opened 2026-05-03
Refactor: remove kubernaut_enrich MCP tool, internalize enrichment into select_workflow


jordigilh:
## Summary

Remove `kubernaut_enrich` as a user-facing MCP tool and make enrichment an internal step within `kubernaut_select_workflow`.

## Rationale

- Enrichment is not a user-facing action — it's a prerequisite for workflow selection
- The LLM needs enrichment data (owner chain, labels, remediation history, spec_hash) to select the correct workflow
- Users don't need to explicitly call enrich; `select_workflow` should do it transparently
- Reduces MCP surface from 3 tools to 2: `kubernaut_investigate` + `kubernaut_select_workflow`

### Remove (mechanical deletions)
- `internal/kubernautagent/mcp/tools/enrich.go` — delete
- `internal/kubernautagent/mcp/tools/registration.go` — remove `EnrichRegistration`
- `cmd/kubernautagent/main.go` — remove `EnrichTool` wiring
- `test/unit/kubernautagent/mcp/tools/enrich_test.go` — delete
- `test/unit/kubernautagent/mcp/tool_registration_test.go` — remove enrich assertions
- `test/unit/kubernautagent/mcp/tools/consistency_test.go` — remove enrich cases
- `test/infrastructure/mcp_e2e_helpers.go` — remove `CallEnrich` helper
- `test/e2e/kubernautagent/interactive_flow_e2e_test.go` — remove enrich step from INT-007
- `test/e2e/kubernautagent/interactive_harm_e2e_test.go` — remove enrich test cases

### Integrate (new work)
- `SelectWorkflowTool.Handle()` calls `Enricher.Enrich()` internally before LLM prompt
- Unit tests: verify enrichment is called within select_workflow
- E2E: validate enrichment data appears in workflow selection LLM context

### Documentation updates
- Update BR-INTERACTIVE.md, INTERACTIVE_MODE_ARCHITECTURE.md, CHANGELOG
- Update test plans (CP-4, CP-5, TEST_PLAN)
- Update proposals (PROPOSAL-EXT-001, PROPOSAL-EXT-003)

## Estimate

~1 day (low risk — enrichment logic in `enricher.go` is unchanged, only the call site moves)

## References

- BR-INTERACTIVE-003 (to be updated)
- DD-INTERACTIVE-002
- kubernaut-apifrontend#3 (MCP tool surface — now 2 proxied tools instead of 3)

jordigilh:
[...]

Issue #891, opened 2026-04-29
KA startup self-check: refuse interactive mode if impersonation SAR fails


jordigilh:
## Context

When `interactive.enabled: true`, KA requires its ServiceAccount to have `impersonate` RBAC (delivered by kubernaut-operator#26). If the RBAC is missing, interactive MCP tool calls will fail at runtime with 403s -- a poor user experience and a silent failure mode.

## Proposal

At KA startup, when `interactive.enabled: true`:

1. Perform a `SelfSubjectAccessReview` to verify the KA SA can impersonate users
2. If the check fails:
   - Log a clear error message: "KA SA lacks impersonate RBAC; interactive mode disabled"
   - Disable interactive mode gracefully (don't register MCP handler)
   - Set a health/readiness condition indicating degraded state
3. If the check passes: proceed normally

This provides:
- **Fail-fast** instead of runtime 403s on first MCP tool call
- **Clear diagnostics** for operators deploying without operator#26
- **Graceful degradation** instead of crash-loop or silent breakage

## Acceptance criteria

- [ ] Startup SAR check for `impersonate` verb on users/groups/serviceaccounts
- [ ] Interactive mode disabled (not crashed) when SAR fails
- [ ] Log message clearly states what's missing and how to fix (reference operator#26)
- [ ] Health endpoint reflects degraded state when interactive mode is force-disabled

## References

- kubernaut-operator#26: KA SA impersonate RBAC delivery
- #703: KA MCP Interactive Mode
- DD-AUTH-MCP-001: Security architecture

jordigilh:
Closed: delivered in PR #1018 (startup SAR self-check for impersonation permissions on both `users` and `groups`, with InteractiveReadiness state tracker and K8s event emission).

PR #1018, opened 2026-05-03
feat(interactive): internalize enrich into select_workflow (#1012) + startup SAR self-check (#891)

jordigilh:
## Summary

Week 1 deliverables for v1.5 interactive mode refinements:

### #1012 — Enrich tool refactor (mechanical)
- **Internalized** `kubernaut_enrich` into `kubernaut_select_workflow`, reducing the MCP tool surface from 3 → 2 tools
- **Introduced `PreSelectionHook` pipeline** abstraction in `SelectWorkflowTool` to prepare for Goose recipe prompt injection points (`pre-workflow-selection` phase, per PROPOSAL-EXT-003)
- `EnrichmentRunner` registered as the first `PreSelectionHook` via `WithEnrichmentRunner` functional option; **autonomous flow unchanged**
- Deleted standalone `enrich.go`, `EnrichRegistration`, `ToolDeps.Enrich`
- Updated E2E tests (`interactive_flow`, `interactive_harm`) to use `select_workflow`
- Updated architecture docs (PROPOSAL-EXT-002, PROPOSAL-EXT-003) to reflect the canonical **five-phase pipeline** model

### #891 — SAR self-check for interactive mode
- **New `internal/kubernautagent/rbac` package** with `CheckImpersonatePermission` (SelfSubjectAccessReview at startup) and `InteractiveReadiness` thread-safe state tracker
- SSAR checks **both `users` and `groups`** resources — interactive mode requires both `Impersonate-User` and `Impersonate-Group` headers
- If the SA **lacks impersonate RBAC** on either resource, interactive mode is **soft-disabled** (autonomous mode continues)
- `/readyz` health endpoint now reports `interactive_mode` (`enabled` | `soft_disabled` | `not_configured`) and `interactive_reason`
- **K8s Events** emitted against the KA pod (via downward API `POD_NAME`/`POD_NAMESPACE`): `InteractiveSoftDisabled` (Warning) or `InteractiveEnabled` (Normal)
- EventEmitter resolves Pod UID for proper event deduplication
- 17 unit tests covering SSAR allowed/denied/error (users, groups, both), readiness states, EventEmitter nil-safety, event content verification, Pod UID lookup

### Architectural note: `post-investigation` phase removal

The pipeline model changes from 6 phases to 5, removing `post-investigation`. **This is intentional**:

- The `post-investigation` hook phase was originally proposed in PROPOSAL-EXT-003 as a placeholder for post-processing after the investigation completes.
- After design review, it was determined that `pre-workflow-selection` **subsumes the use case**: any logic that would run "after investigation" can run as a pre-selection hook, where the full investigation result is already available in the context.
- No downstream consumers or existing designs reference `post-investigation` — it was never implemented or exposed in any CRD.
- Rationale: reducing unnecessary hook surface area simplifies the pipeline model and avoids the ambiguity of having two adjacent optional phases (`post-investigation` and `pre-workflow-selection`) with overlapping data contracts.

The canonical five-phase model is: `pre-investigation` → `investigation` → `rca-resolution` → `pre-workflow-selection` → `workflow-selection`.

## Files changed

| Area | Files | Change |
|------|-------|--------|
| MCP tools | `select_workflow.go`, `enrich.go` (deleted), `registration.go`, `server.go` | Internalize enrich, add pipeline, add WithLogger + debug log |
| KA startup | `main.go` | SSAR check (users+groups), health endpoint, event emitter wiring |
| RBAC | `rbac/impersonate_check.go`, `rbac/event.go` | New package — SSAR, readiness tracker, event emitter with Pod UID |
| Tests | `select_workflow_test.go`, `consistency_test.go`, `tool_registration_test.go`, `enrich_test.go` (deleted), `rbac/*_test.go` | Aligned to 2-tool surface + SSAR + event content verification |
| E2E | `interactive_flow_e2e_test.go`, `interactive_harm_e2e_test.go`, `mcp_e2e_helpers.go` | Updated to use select_workflow |
| Docs | PROPOSAL-EXT-002, PROPOSAL-EXT-003, ADR-KA-001, ROADMAP | Five-phase pipeline, Goose recipe alignment |

## Test plan

- [x] `go build ./...` — full codebase clean
- [x] 29 KA unit test packages pass (`go test ./test/unit/kubernautagent/...`)
- [x] 39 MCP tools unit tests pass
- [x] 17 RBAC unit tests pass (SSAR users+groups, InteractiveReadiness, EventEmitter with UID, event content)
- [x] Pre-commit anti-pattern hooks pass (including NULL-TESTING check)
- [ ] CI pipeline (GitHub Actions)
- [ ] E2E in Kind cluster (interactive flow with/without impersonate RBAC)

## Review feedback addressed

| # | Severity | Issue | Fix |
|---|----------|-------|-----|
| 1 | **BLOCKING** | SSAR only checks `users`, not `groups` | Now checks both; combined denial reasons |
| 2 | **BLOCKING** | NULL-TESTING in DetectPodIdentity test | Sets env vars, asserts both return values |
| 3 | Warning | Event tests only check "doesn't panic" | Verifies `client.Actions()` event content |
| 4 | Warning | Event Pod object lacks UID | Pod GET at emitter creation for UID resolution |
| 5 | Warning | Silent enrichment skip | Debug log (V(1)) + `WithLogger` option |
| 6 | Warning | post-investigation removal undocumented | Rationale added to PR description above |

## Authority

- BR-INTERACTIVE-001..008 (interactive mode business requirements)
- DD-AUTH-MCP-001 (impersonation design decision)
- PROPOSAL-EXT-002, PROPOSAL-EXT-003 (Goose recipe pipeline)
- Issue #1012, Issue #891


internal/kubernautagent/rbac/impersonate_check.go
```diff
@@ -0,0 +1,129 @@
...
+// whether KA's ServiceAccount can impersonate users. This is a prerequisite
+// for interactive mode (DD-AUTH-MCP-001): all MCP tool calls run under the
+// authenticated user's identity via K8s impersonation.
+func CheckImpersonatePermission(ctx context.Context, client kubernetes.Interface) (ImpersonateCheckResult, error) {
+	ssar := &authorizationv1.SelfSubjectAccessReview{
+		Spec: authorizationv1.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &authorizationv1.ResourceAttributes{
+				Verb:     "impersonate",
+				Group:    "",
+				Resource: "users",
```
jordigilh: **BLOCKING**: This SSAR only checks `impersonate` on `users`. Interactive mode also requires impersonating `groups` — the middleware sets both `Impersonate-User` and `Impersonate-Group` headers.

If the SA has `impersonate` on `users` but **not** on `groups`, the self-check passes (interactive mode starts) but runtime impersonation fails with a 403 on the first MCP tool call that tries to set groups.

Suggested fix: add a second SSAR check for `resource: "groups"` and require both to pass:

```go
for _, resource := range []string{"users", "groups"} {
    ssar := &authorizationv1.SelfSubjectAccessReview{
        Spec: authorizationv1.SelfSubjectAccessReviewSpec{
            ResourceAttributes: &authorizationv1.ResourceAttributes{
                Verb:     "impersonate",
                Resource: resource,
            },
        },
    }
    // ... check each, fail if any denied
}
```

test/unit/kubernautagent/rbac/event_test.go
```diff
@@ -0,0 +1,95 @@
...
+		})
+	})
+
+	Describe("DetectPodIdentity", func() {
+
+		It("should return empty strings when env vars are not set", func() {
+			podName, namespace := karbac.DetectPodIdentity()
+			// In a test environment without downward API, these will be empty
+			// unless set by the CI or the test runner.
+			_ = podName
```
jordigilh: **BLOCKING — NULL-TESTING anti-pattern**: This test calls `DetectPodIdentity()` and discards both results with `_ = podName` / `_ = namespace`. There are zero assertions.

Per `.cursor/rules/03-testing-strategy.mdc` §Testing Anti-Patterns:

> ```go
> // ❌ FORBIDDEN: Weak assertions
> Expect(result).ToNot(BeNil())
> ```

This is worse — no assertions at all.

Suggested fix: either assert the empty case and add a positive case with env vars:

```go
It("should return empty strings when env vars are not set", func() {
    os.Unsetenv("POD_NAME")
    os.Unsetenv("POD_NAMESPACE")
    podName, namespace := karbac.DetectPodIdentity()
    Expect(podName).To(BeEmpty())
    Expect(namespace).To(BeEmpty())
})

It("should return env var values when set", func() {
    os.Setenv("POD_NAME", "ka-pod-abc")
    os.Setenv("POD_NAMESPACE", "kubernaut-system")
    defer os.Unsetenv("POD_NAME")
    defer os.Unsetenv("POD_NAMESPACE")
    podName, namespace := karbac.DetectPodIdentity()
    Expect(podName).To(Equal("ka-pod-abc"))
    Expect(namespace).To(Equal("kubernaut-system"))
})
```

test/unit/kubernautagent/rbac/event_test.go
```diff
@@ -0,0 +1,95 @@
...
+			}).NotTo(Panic())
+			Expect(func() {
+				emitter.EmitInteractiveEnabled()
+			}).NotTo(Panic())
+			Expect(func() {
+				emitter.Shutdown()
+			}).NotTo(Panic())
+		})
+
+		It("should emit soft-disabled event without error", func() {
```
jordigilh: **Warning**: This test (and the enabled variant below) only asserts `NotTo(Panic())`. It doesn't verify the event was actually created with the correct `Type`, `Reason`, or `Message`.

The fake clientset records all API calls — you can inspect `client.Actions()` to verify a `create` on `events` with the expected reason and message:

```go
It("should emit a Warning event with InteractiveSoftDisabled reason", func() {
    client := fake.NewSimpleClientset()
    emitter := karbac.NewEventEmitter(client, "ka-pod-abc", "kubernaut-system")
    emitter.EmitInteractiveSoftDisabled("SA lacks impersonate permission")
    emitter.Shutdown()

    actions := client.Actions()
    Expect(actions).NotTo(BeEmpty())
    // Find the create-event action and assert Type/Reason/Message
})
```

Without this, a regression that silently drops events would not be caught.

internal/kubernautagent/rbac/event.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// EmitInteractiveSoftDisabled emits a Warning event when interactive mode
+// is soft-disabled due to missing RBAC (impersonate verb denied).
+func (e *EventEmitter) EmitInteractiveSoftDisabled(reason string) {
+	if e == nil {
+		return
+	}
+	e.recorder.Event(
+		&corev1.Pod{ObjectMeta: metav1.ObjectMeta{
```
jordigilh: **Warning**: This constructs a `corev1.Pod` with `Name` and `Namespace` but **no UID**. In a real cluster, the resulting event will have an empty `involvedObject.uid`.

This causes two issues:
1. Event deduplication becomes unreliable if the pod restarts (same name, different UID)
2. `kubectl describe pod` may not correlate the event correctly

Consider fetching the Pod's UID at `NewEventEmitter` time:

```go
pod, err := clientset.CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
if err == nil {
    e.podRef.UID = pod.UID
}
```

Or pass the Pod UID via downward API (`metadata.uid`) alongside `POD_NAME` and `POD_NAMESPACE`.

internal/kubernautagent/mcp/tools/select_workflow.go
```diff
@@ -40,33 +48,113 @@ type CatalogWorkflow struct {
...
+// SelectWorkflowOption configures optional dependencies on SelectWorkflowTool.
+type SelectWorkflowOption func(*SelectWorkflowTool)
+
+// WithEnrichmentRunner registers enrichment as the first pre-selection hook.
+// Enrichment runs before any Goose recipe prompt injection hooks so that
+// recipe parameters have access to the full enrichment context (#1012).
+func WithEnrichmentRunner(runner EnrichmentRunner) SelectWorkflowOption {
+	return func(t *SelectWorkflowTool) {
+		hook := func(ctx context.Context, input SelectWorkflowInput, user mcpinternal.UserInfo, pctx *PreSelectionContext) error {
+			if input.Kind == "" {
```
jordigilh, line 109: **Warning — silent skip**: When `input.Kind` is empty, enrichment is silently skipped with no log or indication to the caller.

In the old standalone `enrich.go`, `Kind` was a required field with explicit validation error. Now it's optional and silently ignored, which could mask misconfiguration (e.g., the LLM consistently omits `kind` from tool calls and the operator never notices enrichment isn't running).

Suggested: add a debug log so operators can diagnose missing enrichment:

```go
if input.Kind == "" {
    // Consider: logger.V(1).Info("enrichment skipped: kind not provided in input")
    return nil
}
```

Alternatively, document in the MCP tool description that `kind` is required for enrichment to run.

docs/architecture/proposals/PROPOSAL-EXT-003-goose-runtime-evaluation.md
```diff
@@ -44,126 +56,136 @@ This evaluation was refined through two rounds of adversarial audit (14 findings
...
-- **EXT-002 updates deferred** to a follow-up PR so this proposal can stay focused on the narrowed near-term scope: current prompt builder validation first, A2A-only remote hooks second, Goose later.
+- **Goose recipes directly** -- no Kubernaut-native wrapper format. Recipes are the unit of investigation logic, packaged as OCI artifacts for enterprise distribution.
+- **KA as pure orchestrator** -- drops `runLLMLoop`, `llm.Client`, tool registry, and LangChainGo dependency. KA retains pipeline orchestration, CRD lifecycle, signal enrichment, result assembly, permission ceiling enforcement, and audit assembly.
+- **`goose-server` HTTP invocation** -- KA creates Goose sessions via `goose-server`'s HTTP API, which supports recipe-based session creation today. Migration to ACP via `coder/acp-go-sdk` once the upstream session-configuration gap ([goose#7596](https://github.com/block/goose/issues/7596)) closes.
+- **Shadow agent as Goose recipe** -- a parallel Goose session running a security-focused recipe. KA relays `SessionUpdate` events from the primary session to the shadow for real-time prompt injection detection.
```
jordigilh, line 68: **Warning — undocumented architectural change**: This PR changes the pipeline from 6 phases to 5 by removing `post-investigation`. This is a significant architectural decision (it removes a customer hook point) that should be called out in the PR description rather than buried in the doc diffs.

Questions to address:
- Is the removal intentional or accidental?
- If intentional, what's the rationale? (e.g., `pre-workflow-selection` subsumes the use case?)
- Are there any downstream consumers or designs that reference `post-investigation`?

jordigilh (commented):
## Review: PR #1018 — internalize enrich (#1012) + startup SAR (#891)

The code structure is solid and the PreSelectionHook pipeline is a clean abstraction. Two **blocking issues** and four **warnings** flagged inline.

### Blocking (must fix before merge)

1. **SSAR only checks users, not groups** — `impersonate_check.go` line 50. Interactive mode impersonates both users AND groups. If the SA has impersonate on users but not groups, the self-check passes but runtime impersonation fails with 403.

2. **NULL-TESTING anti-pattern** — `event_test.go` line 91. `DetectPodIdentity` test has zero assertions (`_ = podName`). Violates the testing strategy.

### Warnings (should address)

3. **Event emission tests only check "doesn't panic"** — `event_test.go` line 64. Should verify `client.Actions()` contains the expected event type/reason/message.

4. **Event Pod object lacks UID** — `event.go` line 84. Constructed Pod has no UID, causing event deduplication issues in production.

5. **Silent enrichment skip** — `select_workflow.go` line 108. `input.Kind == ""` returns nil with no log. Could mask misconfiguration.

6. **post-investigation phase removal not documented** — PROPOSAL-EXT-003 changes pipeline from 6 to 5 phases. Architectural change should be called out in PR description.

jordigilh (commented):
## Review Round 2: APPROVED (Comprehensive + Adversarial + Security Audit)

All 6 findings from round 1 are addressed:

| Finding | Status |
|---------|--------|
| SSAR only checks `users`, not `groups` | **Fixed** — `impersonateResources = []string{"users", "groups"}` with loop, combined denial reasons |
| NULL-TESTING `DetectPodIdentity` | **Fixed** — positive test with `os.Setenv` + assertions, negative with `Expect(BeEmpty())` |
| Event tests only check "doesn't panic" | **Fixed** — `Eventually` + `filterActions` + asserts `Reason`, `Type`, `Message`, `InvolvedObject.Name` |
| Event Pod lacks UID | **Fixed** — `Pods().Get()` with 5s timeout, sets `podObj.UID`, graceful fallback |
| Silent enrichment skip | **Fixed** — `t.logger.V(1).Info(...)` + `logr.Logger` field + `WithLogger` option + default `logr.Discard()` |
| `post-investigation` removal undocumented | Acknowledged — PR description concern, not a code issue |

### Adversarial Security Audit (10 attack vectors reviewed)

| ID | Vector | Risk | Verdict |
|----|--------|------|---------|
| A1 | SSAR TOCTOU race (RBAC changes after startup) | Low | Acceptable — runtime 403 is the real guard; self-check is early warning |
| A2 | Missing `uids` in SSAR check (KEP-1513) | Low | Correct — KA never sets `Impersonate-Uid` |
| A3 | Info disclosure in event messages | Low | Events are namespace-scoped, same RBAC as pod access |
| A4 | Pod UID lookup as namespace probe | Very Low | `POD_NAME`/`POD_NAMESPACE` from downward API, not external input |
| A5 | 5s timeout on UID lookup at startup | Low | Bounded, graceful fallback, before readiness probes |
| A6 | Enrichment impersonation context injection | Secure | Uses middleware-authenticated `UserInfo`, not untrusted input |
| A7 | Hook ordering guarantees (no priority system) | Low | Convention-based (enrichment first by caller ordering); acceptable for v1.5 |
| A8 | Error message leak in enrichment errors | Low | MCP client is authenticated K8s user, no additional info exposed |
| A9 | `InteractiveReadiness` thread safety | Verified | `sync.RWMutex` on all read/write paths |
| A10 | Event UID propagation not tested | Minor | GET verified but UID not asserted on event object — non-blocking |

### Merge Notes

- **Doc conflicts expected**: PROPOSAL-EXT-003, ADR-KA-001, shadow-agent-configuration.md, and ROADMAP.md were independently modified on `development/v1.5` (Goose deferral to v1.6). Merge will require conflict resolution on those 4 files.
- **E2E pending**: PR checklist shows E2E in Kind not yet validated. Should be run before final merge.

**Confidence: 92%** — All blocking issues fixed, security model sound, behavioral test coverage verified. Remaining gap: E2E validation + doc merge conflicts.

LGTM — ready to merge once E2E passes and doc conflicts are resolved.

PR #1035, opened 2026-05-05
fix(gateway): dynamic owner resolution + FedRAMP readiness remediation (#1029, #1032)

### LINKED ISSUE #1029: Gateway owner-resolver: add dynamic API discovery, Service-to-workload traversal, and cross-namespace resolution
## Problem

The Gateway's owner-resolver uses a static list of known Kubernetes resource kinds (Deployment, StatefulSet, DaemonSet, Pod, etc.) when resolving the `targetResource` for incoming Prometheus alert signals. OpenShift CRD kinds such as `BuildConfig`, `DeploymentConfig`, and `Route` are not recognized, causing the RR to be created with `targetResource.kind: Unknown` and `targetResource.name: unknown`.

This was discovered during v1.4 scenario validation on OCP. Four distinct failure modes have been identified:

---

### Use Case 1: `build-failure` — CRD kind not in static map

The `BuildFailureRate` alert carries a `buildconfig: webapp` label (injected by `openshift_build_status_phase_total`), but the Gateway cannot resolve it to `BuildConfig/webapp`.

**Observed behavior:**
1. Alert `BuildFailureRate` fires with labels `{buildconfig: "webapp", namespace: "demo-build"}`
2. Gateway receives the signal but cannot resolve the owner — sets `targetResource: {kind: Unknown, name: unknown}`
3. AI Analysis correctly identifies `Build/webapp-2` as the failing resource and the root cause (invalid Git source URI)
4. Workflow resolution fails with `rca_incomplete` because no workflow matches target kind `Build`; the `fix-build-source-v1` workflow expects `BuildConfig`
5. RR completes as `ManualReviewRequired` instead of being auto-remediated

### Use Case 2: `route-misconfiguration` — cross-namespace metric source resolution

The `RouteBackendUnavailable` alert is sourced from HAProxy router metrics (`haproxy_backend_up`). The metric labels include `service: "router-internal-default"` (the Prometheus scrape target — the OCP router Service in `openshift-ingress`) and `exported_namespace: "demo-route"` (the actual Route's namespace).

**Observed behavior:**
1. Alert `RouteBackendUnavailable` fires with labels `{service: "router-internal-default", namespace: "demo-route", exported_namespace: "demo-route", route: "storefront"}`
2. Gateway's owner-resolver extracts `service: "router-internal-default"` and attempts to resolve `Service/router-internal-default` in namespace `demo-route`
3. That service does not exist in `demo-route` — it lives in `openshift-ingress` (it's the OCP router's own service)
4. Owner resolution fails: `services "router-internal-default" not found`
5. All alerts in the batch fail and the signal is dropped entirely — **no RR is ever created**

**Key difference from Use Case 1:** In this case, the RR is never created at all (the signal is dropped before it reaches the operator), whereas in Use Case 1 the RR is created but with an unresolvable target kind.

### Use Case 3: `operator-health` — OLM Subscription/CSV not resolvable

The `OperatorCSVFailed` alert uses `absent()` to detect when a `csv_succeeded` metric disappears (CSV deleted). The alert carries static labels `target_kind: Subscription`, `target_name: etcd`, `operator_name: etcdoperator`, but the Gateway cannot resolve `Subscription` (an OLM CRD from `operators.coreos.com/v1alpha1`).

**Observed behavior:**
1. Alert `OperatorCSVFailed` fires with labels `{namespace: "demo-operator", target_kind: "Subscription", target_name: "etcd", operator_name: "etcdoperator"}`
2. Gateway's owner-resolver ignores the custom `target_kind`/`target_name` labels entirely — they are not in the static label-to-kind map
3. No other resolvable label is present (`absent()` strips all metric labels)
4. RR is created with `targetResource: {kind: Unknown, name: unknown, namespace: demo-operator}`
5. AI Analysis enters `Investigating` phase but has no target resource context — spends 5+ minutes making tool calls blind
6. Investigation fails with `rca_incomplete` → `ManualReviewRequired`

**Key difference:** Unlike Use Cases 1-2 where the alert payload contains standard metric labels that could theoretically be mapped, here the `absent()` function strips all labels. The only viable path is for the Gateway to honor `target_kind`/`target_name` labels from the alert payload as explicit target hints.

**Additional note:** Even if the Gateway did resolve `Subscription/etcd`, it would need to understand the OLM ownership chain: `Subscription` → `InstallPlan` → `ClusterServiceVersion`. The AI needs to investigate the CSV, not the Subscription itself.

### Use Case 4: `etcd-defrag-forecast` — Service resolved instead of StatefulSet

The `EtcdHighFragmentationRatio` alert fires with a `service: "etcd-client"` label from the ServiceMonitor scrape. The Gateway resolves `Service/etcd-client` and stops there, rather than traversing to the owning `StatefulSet/etcd`.

**Observed behavior:**
1. Alert `EtcdHighFragmentationRatio` fires with labels `{service: "etcd-client", namespace: "demo-etcd-defrag"}`
2. Gateway resolves `Service/etcd-client` — this Service exists in the namespace
3. However, the Service is not the workload — it's a headless Service fronting `StatefulSet/etcd`
4. RR is created with `targetResource: {kind: Service, name: etcd-client}` instead of `{kind: StatefulSet, name: etcd}`
5. AI Analysis identifies the StatefulSet during investigation but cannot match it to a `defrag-etcd-v1` workflow that expects StatefulSet targets
6. RR completes as `ManualReviewRequired`

**Expected behavior:** The Gateway should traverse the Service → Endpoints → Pod → ownerRef chain to discover `StatefulSet/etcd` as the ultimate workload owner.

---

### Expected behavior

The Gateway should use the Kubernetes dynamic API (discovery client) to resolve resource kinds from alert labels, rather than relying on a hardcoded list. Additionally:

1. **Dynamic CRD resolution** — discover that `BuildConfig`, `Route`, `Subscription`, etc. are valid API resources via discovery
2. **Cross-namespace awareness** — when `exported_namespace` is present, prefer it for determining the resource's home namespace
3. **Service-to-workload traversal** — when resolving a `Service`, traverse the Endpoints → Pod → ownerRef chain to find the ultimate workload (Deployment, StatefulSet, DaemonSet)
4. **Honor explicit target hints** — when alert labels include `target_kind`/`target_name`, use them as authoritative hints for the target resource (falling back to discovery-based resolution)

For Use Case 1 (BuildConfig):
1. Discover that `BuildConfig` is a valid API resource (via `build.openshift.io/v1`)
2. Resolve `targetResource: {kind: BuildConfig, name: webapp, namespace: demo-build}`
3. Allow the AI analysis to target `BuildConfig/webapp` directly, matching the `fix-build-source-v1` workflow

For Use Case 2 (Route/HAProxy):
1. Recognize that `exported_namespace` indicates the alert pertains to a resource in a different namespace than the metric source
2. Use the `route` label to resolve `Route/storefront` in `demo-route` (via `route.openshift.io/v1`)
3. Do NOT attempt to resolve `Service/router-internal-default` in `demo-route` — that service belongs to the infrastructure namespace `openshift-ingress`

For Use Case 3 (Subscription/CSV):
1. Honor `target_kind: Subscription` / `target_name: etcd` from alert labels as explicit target hints
2. Resolve `targetResource: {kind: Subscription, name: etcd, namespace: demo-operator}` (via `operators.coreos.com/v1alpha1`)
3. Allow the AI to investigate the Subscription → InstallPlan → CSV chain with proper context

For Use Case 4 (Service → StatefulSet):
1. When resolving `Service/etcd-client`, discover it is a headless Service
2. Traverse Endpoints → Pod → ownerRef to find `StatefulSet/etcd`
3. Resolve `targetResource: {kind: StatefulSet, name: etcd, namespace: demo-etcd-defrag}`

### Root cause

`pkg/gateway/adapters/owner_resolver.go` maps alert labels to resource kinds using a static map. Labels like `deployment`, `statefulset`, `daemonset` are handled, but `buildconfig`, `deploymentconfig`, `route`, `subscription`, and other CRD labels are not.

Furthermore:
- The resolver assumes that the `service` label always refers to an application Service in the alert's `namespace`. For infrastructure-sourced metrics like HAProxy, the `service` label refers to the Prometheus scrape target, not an application resource.
- The resolver does not traverse Service → Endpoints → Pod → ownerRef to discover workload owners.
- Custom alert labels (`target_kind`, `target_name`) are ignored during owner resolution.

### Suggested fix

Replace the static kind map with a dynamic approach:

1. Use the discovery client (`client-go/discovery`) to enumerate all API resources at startup (and periodically refresh)
2. Build a label-to-GVK mapping from the discovered resources (e.g., `buildconfig` → `build.openshift.io/v1/BuildConfig`, `route` → `route.openshift.io/v1/Route`, `subscription` → `operators.coreos.com/v1alpha1/Subscription`)
3. Fall back to the static map for well-known kinds to avoid regression
4. Add awareness of `exported_namespace` — when present, prefer it over `namespace` for determining the resource's home namespace
5. Add Service → workload traversal: when the resolved resource is a Service, follow Endpoints → Pod → ownerRef to the workload controller
6. Honor explicit `target_kind`/`target_name` labels in alert payloads as authoritative target hints

### Impact

| Scenario | Status | Failure Mode |
|---|---|---|
| `build-failure` | BLOCKED | RR created but unresolvable target → `ManualReviewRequired` |
| `route-misconfiguration` | BLOCKED | Signal dropped entirely → no RR created |
| `operator-health` | BLOCKED | RR created with Unknown target, AI blind → `ManualReviewRequired` |
| `etcd-defrag-forecast` | BLOCKED | RR targets Service instead of StatefulSet → `ManualReviewRequired` |
| Any future OCP CRD scenario | BLOCKED | Same pattern — static map does not cover CRD kinds |
| Vanilla Kubernetes | NOT AFFECTED | No CRD kinds or cross-namespace metrics in alert labels |

### Workaround

None currently viable. Adding `target_kind` / `target_name` labels to the PrometheusRule alert does not help because the Gateway ignores custom labels during owner resolution.

### Environment

- Kubernaut: v1.4.0-rc5
- Platform: OpenShift 4.x (OCP)
- Gateway version: latest (as of 2026-05-04)

### LINKED ISSUE #1032: Gateway silently drops entire alert batch when one alert fails owner resolution
## Problem

When AlertManager sends a batch of grouped alerts to the Gateway webhook, if **any single alert** in the batch fails during owner resolution (signal processing), the **entire batch is rejected** and all alerts are silently dropped. No RemediationRequest is created for any alert in the batch, including those that would have resolved successfully.

## Observed behavior

During v1.4.0-rc5 OCP validation:

1. AlertManager grouped `SCCViolationPodBlocked` (demo-scc) and `RBACPolicyDenied` (demo-rbac) into a single webhook POST (default grouping by `namespace=~"^demo-.*$"`)
2. The `RBACPolicyDenied` alert's owner resolution failed at the Gateway (unrelated issue in the workflow)
3. The **entire batch** was rejected — including the perfectly valid `SCCViolationPodBlocked` alert
4. No RR was created for either alert
5. No error was surfaced to indicate the SCC alert was dropped — **silent signal loss**

The Gateway logs showed `AlertmanagerFailedToSendAlerts` errors for the batch, but no per-alert failure detail.

## Expected behavior

The Gateway should process each alert in a batch **independently**. If one alert fails owner resolution:
- Other alerts in the batch should still be processed and create RRs
- The failed alert should be logged with its specific error
- A metric or event should track per-alert failures for observability

## Workaround

Configure AlertManager with `group_by: [alertname, namespace]` on the Gateway webhook route to ensure each alert type is sent in its own batch:

```yaml
route:
  receiver: gateway-webhook
  group_by: [alertname, namespace]
  match_re:
    namespace: "^demo-.*$"
```

This is fragile because:
- Customers must know to configure this
- Default AlertManager grouping behavior will trigger the bug
- It doesn't prevent the issue if two alerts of the same `alertname` in the same `namespace` have different resolution outcomes

## Impact

- **Silent signal loss** — valid alerts are dropped with no indication in the RR lifecycle
- **Non-deterministic** — whether an alert is processed depends on what other alerts happen to be in the same AlertManager batch
- **Hard to debug** — no per-alert error trail; the only symptom is "expected RR never appeared"

## Environment

- Kubernaut: v1.4.0-rc5
- Platform: OpenShift 4.x
- AlertManager: default OCP cluster monitoring stack

### LINKED ISSUE #1036: Gateway creates only 1 RR when AlertManager batch contains multiple alerts with distinct owner chains
## Problem

When AlertManager sends a grouped webhook batch containing multiple alerts that resolve to **different** owner chains, the Gateway creates only 1 RemediationRequest instead of N.

This was discovered during the `cascading-service-failure` scenario validation on OCP.

## Scenario

The `cascading-service-failure` scenario has 3 Deployments in `demo-cascade`:
- `postgres` (root cause — injected crash)
- `order-processor` (dependent — CrashLoopBackOff when postgres is down)
- `inventory-sync` (dependent — CrashLoopBackOff when postgres is down)

When both `order-processor` and `inventory-sync` enter CrashLoopBackOff, `KubePodCrashLooping` fires for each pod. These are distinct pods owned by distinct Deployments.

## Observed behavior

1. AlertManager groups both alerts by `[alertname, namespace]` → sends a single webhook payload with 2 alerts in the batch
2. Each alert has a different `pod` label (`order-processor-xxx` and `inventory-sync-xxx`)
3. Owner resolution should produce two distinct fingerprints:
   - `demo-cascade:Deployment:order-processor`
   - `demo-cascade:Deployment:inventory-sync`
4. **Only 1 RR is created** — the second alert in the batch is effectively dropped

## Expected behavior

The Gateway should iterate over all alerts in the batch, resolve each to its owner chain independently, and create a separate RR for each distinct fingerprint (`namespace:kind:name`). In this case, 2 RRs should be created.

## Root cause hypothesis

The Gateway may be:
- Processing only the first alert in a batch and short-circuiting, OR
- Deduplicating within the batch before resolving individual owner chains, OR
- Using the batch-level grouping labels (which are identical for both alerts) as the fingerprint key instead of the per-alert owner chain

## Workaround

Adding `pod` to the AlertManager `group_by` (e.g., `group_by: [alertname, namespace, pod]`) forces AlertManager to send separate webhooks per pod. However, this is not a proper fix since:
- It increases webhook volume significantly
- Other grouped alerts (e.g., multiple replica failures for the same Deployment) would no longer be batched correctly
- The Gateway should handle multi-alert batches correctly regardless of grouping

## Impact

| Scenario | Effect |
|---|---|
| `cascading-service-failure` | Only 1 of 2 dependent Deployments gets an RR |
| Any multi-workload failure in same namespace | Missing RRs for subsequent workloads in the batch |

## Environment

- Kubernaut: v1.4.0-rc5
- Platform: OpenShift 4.x (OCP)
- AlertManager grouping: `group_by: [alertname, namespace]`

jordigilh:
## Summary

- **Dynamic API Resource Registry (#1029)**: Replace static `kindToGroup` and `resourceCandidates` maps with a fully dynamic `APIResourceRegistry` backed by Kubernetes API discovery. Multi-candidate scoring with tier-based priority, deterministic tie-breaking, and real existence validation via dynamic client GET. Panic-resilient refresh loop with capped re-entry.

- **Service→Workload Traversal (#1029 Use Case 3)**: When the owner resolver encounters a Service target (from Prometheus `service` scrape labels), traverse `Service → spec.selector → Pods → ownerRefs` to resolve the actual backing workload (e.g., `StatefulSet/etcd`). Fixes the `etcd-defrag-forecast` scenario where `EtcdHighFragmentationRatio` resolved to `Service/etcd-client` instead of `StatefulSet/etcd`.

- **Batch-Independent Alert Processing (#1032, #1036)**: Wire `ParseBatch` end-to-end via optional `BatchParser` interface (type assertion — `SignalAdapter` untouched). Handler returns HTTP 207 Multi-Status with per-alert results. Each signal calls `ProcessSignal` independently, preserving per-fingerprint deduplication. This also fixes #1036 where a grouped AlertManager batch with alerts from different owner chains (e.g., `cascading-service-failure` scenario) produced only 1 RR instead of N — `ParseBatch` iterates all alerts independently while `Parse` short-circuits on the first success.

- **FedRAMP Security Hardening**: Enrich audit events with authenticated K8s identity (AU-3), sanitize validation error messages, document tamper-evidence delegation (SC-8/AU-9), fail startup on invalid TLS profile.

- **Operations Fixes**: Remove duplicate histogram observations (3→1), remove dead `InFlightRequests` middleware, context-aware graceful shutdown, fix stale docstrings.

- **Dev Manifest Cleanup**: Align `deploy/gateway/` Kustomize manifests with operator (remove `secrets`, add `leases`, fix `apiGroups`, update probes).

- **QE Improvements**: Add `--race` to all unit/integration Makefile test targets, replace `time.Sleep` with `Eventually` in tests, delete disconnected `label_filter.go`, reconcile `TEST_PLAN.md`.

- **PR Review Fixes (P1–P3)**: Wire `OwnerResolutionTotal` and `SignalsParseDroppedTotal` metrics in production code, add RFC 1123 validation for alert label values before K8s API calls, cap existence cache (10k) with LRU eviction, use `singleflight.Group` for CheckExistence, add discovery refresh error counter, remove debug log artifacts, move test helpers to `testing_exports.go`, document HTTP 207 JSON variant.

## Commits

1. `feat(gateway): dynamic API resource registry with existence validation (#1029)` — core registry, scoring, metrics, tests
2. `feat(gateway): batch-independent alert processing with FedRAMP hardening (#1032)` — BatchParser, 207 handler, security/ops fixes
3. `chore(gateway): align dev manifests with operator for consistency` — YAML-only, dev/test path
4. `test(gateway): QE improvements and dead code cleanup` — Makefile, test improvements, label_filter removal
5. `feat(gateway): Service→workload traversal for Prometheus scrape labels (#1029 UC3)` — owner resolver Service→selector→Pods→ownerRefs, ResolveFingerprint returns resolved resource
6. `fix(gateway): address all PR #1035 review findings (P1–P3)` — observability, security, resilience, code hygiene
7. `fix(gateway): clarify RFC 1123 lowercase-only intent on label validation` — doc comment nit
8. `test(gateway): add cascading-failure multi-owner batch tests (#1036)` — verifies ParseBatch produces N signals for N distinct owner chains

## Test plan

- [x] `go vet ./pkg/gateway/... ./cmd/gateway/...` — clean
- [x] Unit tests pass with `--race` (514 specs across 7 suites, 0 failures)
- [x] Pre-commit anti-pattern hooks pass on all 8 commits
- [x] Service→workload traversal: 7 test scenarios (UT-GW-1029-041..047)
- [x] Cascading-failure multi-owner batch: 3 test scenarios (UT-GW-1036-001..003)
- [ ] Integration tests (`make test-integration-gateway`)
- [ ] Verify HTTP 207 response for multi-alert Prometheus webhook
- [ ] Verify Service/etcd-client → StatefulSet/etcd resolution end-to-end

Closes #1029, closes #1032, closes #1036


pkg/gateway/metrics/metrics.go
```diff
@@ -82,6 +88,12 @@ type Metrics struct {
 	// Scope Filtering Metrics (BR-SCOPE-002: Gateway Signal Filtering)
 	SignalsRejectedTotal *prometheus.CounterVec // gateway_signals_rejected_total{reason}
 
+	// Owner Resolution Metrics (#1029)
+	OwnerResolutionTotal *prometheus.CounterVec // gateway_owner_resolution_total{kind, outcome}
```
jordigilh, line 95: **P1-OBS-1: `OwnerResolutionTotal` metric registered but never incremented from production code.**

This counter is defined here with labels `{kind, outcome}` but has zero callers outside of `observability_test.go`. No production code path in `owner_resolver.go`, `prometheus_adapter.go`, or `server.go` calls `.OwnerResolutionTotal.WithLabelValues(…).Inc()`.

An unincremented metric in the Prometheus registry misleads SRE dashboards (permanent zeros) and implies a capability that doesn't actually function.

**Fix:** Wire `.OwnerResolutionTotal.WithLabelValues(kind, "success").Inc()` from the owner resolution success path and `(kind, "failed")` / `(kind, "skipped_crd")` from the failure/skip paths.

pkg/gateway/metrics/metrics.go
```diff
@@ -82,6 +88,12 @@ type Metrics struct {
...
+
+	// Batch Parse Drop Metrics (#1032)
+	SignalsParseDroppedTotal *prometheus.CounterVec // gateway_signals_parse_dropped_total{reason}
```
jordigilh, line 98: **P1-OBS-2: `SignalsParseDroppedTotal` metric registered but never incremented from production code.**

This counter is defined here with label `{reason}` but has zero callers outside of `observability_test.go`. Specifically, `PrometheusAdapter.ParseBatch` (prometheus_adapter.go:244-247) skips failed alerts with a log line but emits no drop metric.

**Fix:** Add `metrics.SignalsParseDroppedTotal.WithLabelValues("owner_resolution_failed").Inc()` in the `ParseBatch` skip path, and similarly in `handleBatchRequest` for validation rejections.

pkg/gateway/adapters/resource_registry.go
```diff
@@ -0,0 +1,430 @@
...
+	}()
+
+	ticker := time.NewTicker(r.refreshInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return true
+		case <-ticker.C:
+			_ = r.Refresh(ctx)
```
jordigilh: **P3-OBS-4 (nit): `_ = r.Refresh(ctx)` swallows the error return.**

`Refresh` already logs the error internally and preserves the previous snapshot on failure, so the `_ =` is intentionally non-fatal. However, a counter metric for discovery refresh failures would improve operational visibility (e.g., `gateway_discovery_refresh_errors_total`).

No action required — just a hardening suggestion.

pkg/gateway/adapters/prometheus_adapter.go
```diff
@@ -300,40 +363,85 @@ var resourceCandidates = []resourceCandidate{
...
+			return candidates[i].kind < candidates[j].kind
+		})
+
+		// Return the first candidate that actually exists (if existence checking is available)
+		for _, c := range candidates {
+			gvr, ok := registry.KindToGVR(c.kind)
+			if !ok {
+				continue
+			}
+			if registry.CheckExistence(ctx, gvr, namespace, c.name) {
```
jordigilh, line 451: **P2-ADV-1: Alert label values used as K8s resource names without validation.**

In `extractTargetResource`, label values from Prometheus alerts are passed directly to `registry.CheckExistence(ctx, gvr, namespace, c.name)` → `dynClient.Resource(gvr).Namespace(namespace).Get(ctx, name, …)`. While K8s will reject invalid names server-side, an adversarial alert payload with a 253+ character label value or special characters will generate expensive API calls and pollute logs before failing.

**Fix:** Validate `name` and `namespace` against K8s naming rules (RFC 1123 subdomain, max 253 chars) before making the API call.

pkg/gateway/adapters/prometheus_adapter.go
```diff
@@ -300,40 +363,85 @@ var resourceCandidates = []resourceCandidate{
...
 // 3. default → "default"
 //
-// Returns:
-// - string: Namespace name or "default"
+// In federated setups, the "namespace" label often points to the federation
+// namespace (e.g., "monitoring"), while "exported_namespace" contains the
+// actual workload namespace (e.g., "production").
 func extractNamespace(labels map[string]string) string {
-	if ns, ok := labels["namespace"]; ok {
+	if ns, ok := labels["exported_namespace"]; ok && ns != "" {
```
jordigilh: **P2-ADV-2: `exported_namespace` label trusted directly without validation.**

The `exported_namespace` label from federated Prometheus is used as the K8s namespace for owner resolution and fingerprint generation without any validation. An adversarial AlertManager payload could set this to an unexpected namespace.

While BR-SCOPE-002 scope filtering applies downstream, the fingerprint and owner resolution happen with the unvalidated namespace first.

**Fix:** Validate the namespace against K8s naming rules (RFC 1123 label, max 63 chars) and optionally against the scope checker's managed namespaces.

pkg/gateway/adapters/resource_registry.go
```diff
@@ -0,0 +1,430 @@
...
+// static resourceCandidates and kindToGroup maps with fully dynamic API
+// discovery. Issue #1029.
+type APIResourceRegistry struct {
+	dc        discovery.DiscoveryInterface
+	dynClient dynamic.Interface
+	mu        sync.RWMutex
+	snapshot  *registrySnapshot
+
+	cacheMu  sync.Mutex
+	cache    map[string]existenceCacheEntry
```
jordigilh: **P2-RES-1: `cache map[string]existenceCacheEntry` has no eviction or max-size bound.**

This cache grows with every unique `(gvr, namespace, name)` tuple. The 30s TTL prevents stale reads but entries are never evicted — only overwritten on re-check or bulk-cleared on `Refresh` (every 5 min). In a cluster with many distinct resources and frequent alerts, this map grows linearly between refresh cycles.

**Fix:** Add a max-size cap (e.g., 10,000 entries) with LRU eviction, or add periodic pruning of expired entries.

pkg/gateway/adapters/resource_registry.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// CheckExistence verifies whether a resource exists in the given namespace.
+// Uses a short-lived TTL cache to bound API server load.
+// Returns true if the resource exists, false otherwise.
+// Errors (403, timeout) are treated as "does not exist" with a warning log.
+func (r *APIResourceRegistry) CheckExistence(ctx context.Context, gvr schema.GroupVersionResource, namespace, name string) bool {
+	key := existenceCacheKey(gvr, namespace, name)
+
+	r.cacheMu.Lock()
```
jordigilh, line 417: **P3-CON-1 (nit): `CheckExistence` cache lock pattern allows thundering herd.**

The lock is released after the cache miss check (line 389) and re-acquired for the write (line 411). Between these two points, N concurrent callers for the same key all hit the K8s API. The TTL mitigates impact and the GET is idempotent, so this is not a correctness issue — only a performance concern under burst.

**Suggestion:** A `singleflight.Group` would collapse concurrent lookups for the same key. No action required for this PR.

Makefile
```diff
@@ -244,7 +244,7 @@ test-unit-gateway: ginkgo ensure-coverage-dirs ## Run gateway unit tests (coverp
 	@echo "════════════════════════════════════════════════════════════════════════"
 	@echo "🧪 gateway - Unit Tests ($(TEST_PROCS) procs)"
 	@echo "════════════════════════════════════════════════════════════════════════"
-	@$(GINKGO) -v --timeout=$(TEST_TIMEOUT_UNIT) --procs=$(TEST_PROCS) --coverprofile=coverage_unit_gateway.out --covermode=atomic --coverpkg=github.com/jordigilh/kubernaut/pkg/gateway/... ./test/unit/gateway/...
+	@$(GINKGO) -v --race --timeout=$(TEST_TIMEOUT_UNIT) --procs=$(TEST_PROCS) --coverprofile=coverage_unit_gateway.out --covermode=atomic --coverpkg=github.com/jordigilh/kubernaut/pkg/gateway/... ./test/unit/gateway/...
```
jordigilh, line 247: **P2-CON-2: `--race` added to some test targets but not all.**

The PR adds `--race` to `test-unit-gateway`, `test-unit-kubernautagent`, and `test-integration-%` but not to other unit test targets like `test-unit-notification`, `test-unit-remediationorchestrator`, etc.

The commit message says "Add `--race` to Makefile test targets" (plural, generic) which implies universal coverage.

**Fix:** Either add `--race` to all test targets for consistency, or scope the commit message to clarify gateway/KA only.

pkg/gateway/server.go
```diff
@@ -1044,6 +1052,127 @@ func (s *Server) createAdapterHandler(adapter adapters.SignalAdapter) http.Handl
...
+		}
+
+		if valErr := adapter.Validate(signal); valErr != nil {
+			logger.Info("Signal validation failed in batch",
+				"fingerprint", signal.Fingerprint,
+				"error", valErr)
+			results = append(results, ProcessingResult{
+				Status:      "rejected",
+				Fingerprint: signal.Fingerprint,
+				Error:       "Signal validation failed",
```
jordigilh: **P2-ERR-1: Batch validation failure error could be more actionable.**

The log at line 1121 correctly includes the fingerprint and error, but the HTTP response error field only says `"Signal validation failed"`. The fingerprint IS already present as a separate field (line 1126), so correlation is possible. Minor — just noting for completeness.

pkg/gateway/server.go
```diff
@@ -1044,6 +1052,127 @@ func (s *Server) createAdapterHandler(adapter adapters.SignalAdapter) http.Handl
...
+			summary.Created++
+		}
+	}
+
+	resp := BatchProcessingResponse{
+		Results: results,
+		Summary: summary,
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusMultiStatus)
```
jordigilh, line 1217: **P2-SPEC-1: HTTP 207 Multi-Status JSON response is not RFC 4918 XML.**

HTTP 207 (WebDAV spec, RFC 4918) defines an XML `multistatus` response body. This implementation returns a pragmatic JSON variant, which is common in modern REST APIs but technically deviates from the RFC.

**Fix:** Add a GoDoc comment on `BatchProcessingResponse` and/or the OpenAPI spec noting this is a JSON variant of HTTP 207, not RFC 4918 XML.

pkg/gateway/adapters/prometheus_adapter.go
```diff
@@ -208,6 +211,73 @@ func (a *PrometheusAdapter) Parse(ctx context.Context, rawData []byte) (*types.N
...
+		kind, name := extractTargetResource(ctx, alert.Labels, ns, a.registry)
+		resource := types.ResourceIdentifier{
+			Kind:      kind,
+			Name:      name,
+			Namespace: ns,
+		}
+
+		fingerprint, err := types.ResolveFingerprint(ctx, a.ownerResolver, resource, a.logger)
+		if err != nil {
+			a.logger.Info("Alert failed owner resolution in batch, skipping",
```
jordigilh, line 274: **P1-OBS-2 (related): This is where `SignalsParseDroppedTotal` should be incremented.**

When an alert fails owner resolution and is skipped (`continue` at line 248), the drop is only logged — no metric is emitted. The `gateway_signals_parse_dropped_total{reason="owner_resolution_failed"}` counter registered in `metrics.go` should be incremented here.

Note: The adapter doesn't currently have access to the `Metrics` instance. You'll need to either pass it via constructor or use a callback pattern.

jordigilh (commented):
## 9-Category PR Audit — Verdict: **request-changes**

| Category | P1 | P2 | P3 |
|---|---|---|---|
| 1. Observability wiring | 2 | 1 | 1 |
| 2. Adversarial inputs | 0 | 2 | 1 |
| 3. Resource bounds | 0 | 1 | 0 |
| 4. Concurrency | 0 | 1 | 1 |
| 5. Nil/zero edge cases | 0 | 0 | 0 |
| 6. Error-path observability | 0 | 1 | 0 |
| 7. Cross-component integration | 0 | 0 | 0 |
| 8. Spec compliance | 0 | 1 | 1 |
| 9. API surface hygiene | 0 | 2 | 0 |
| **Total** | **2** | **9** | **4** |

### Minimum to unblock
1. Wire `OwnerResolutionTotal.WithLabelValues(kind, outcome).Inc()` from the owner resolution path (P1-OBS-1)
2. Wire `SignalsParseDroppedTotal.WithLabelValues(reason).Inc()` from the `ParseBatch` skip path (P1-OBS-2)
3. Remove `[DEBUG]` log statements from `emitCRDCreatedAudit` (P2-OBS-3) — lines 2060-2072 in server.go are debug artifacts that fire at `Info` level on every CRD creation

### Additional P2 items (can track as follow-ups)
- **P2-API-1/2**: `SetCacheReadyForTesting`, `SetShuttingDownForTesting`, `NewMinimalServerForReadinessTest` are exported from production package `pkg/gateway` but only referenced from test files. Move to `_test.go` or `internal/testutil`.
- **P2-SPEC-1**: HTTP 207 JSON response should document that it's a JSON variant, not RFC 4918 XML.
- **P2-CON-2**: `--race` added to gateway/KA targets but not other services — either add to all or scope the commit message.

See inline comments for each finding with file + line citations.

jordigilh (commented):
## Re-review: All P1 and P2 findings addressed

Following up on the [previous review](https://github.com/jordigilh/kubernaut/pull/1035#pullrequestreview-4225179342). The team addressed all findings in commit `ccd168a` ("fix(gateway): address all PR #1035 review findings (P1–P3)") plus an additional feature commit `8f5c9a2` (Service→workload traversal).

### Finding Resolution Status

| ID | Severity | Finding | Status |
|---|---|---|---|
| P1-OBS-1 | P1 | `OwnerResolutionTotal` never incremented | ✅ **Fixed** — `SetOwnerResolutionMetric` setter wired from `main.go`; incremented on success/fail in both `Parse` and `ParseBatch` |
| P1-OBS-2 | P1 | `SignalsParseDroppedTotal` never incremented | ✅ **Fixed** — `SetParseDroppedMetric` setter wired from `main.go`; incremented with `owner_resolution_failed` in `ParseBatch` skip path |
| P2-OBS-3 | P2 | `[DEBUG]` log statements in `emitCRDCreatedAudit` | ✅ **Fixed** — Both `[DEBUG]` blocks removed from `server.go` |
| P3-OBS-4 | P3 | `_ = r.Refresh(ctx)` swallows error | ✅ **Fixed** — Error now checked; `refreshErrorCounter.Inc()` called on failure. New `DiscoveryRefreshErrorsTotal` metric added |
| P2-ADV-1 | P2 | Label values used as K8s names without validation | ✅ **Fixed** — `isValidK8sName()` with RFC 1123 regex applied to label values in `extractTargetResource` |
| P2-ADV-2 | P2 | `exported_namespace` trusted without validation | ✅ **Fixed** — Both `exported_namespace` and `namespace` labels validated with `isValidK8sName()` before use in `extractNamespace` |
| P2-RES-1 | P2 | Existence cache unbounded | ✅ **Fixed** — `DefaultMaxCacheSize = 10000` with `evictExpiredLocked()` that prunes expired entries first, then resets if still over cap |
| P2-CON-2 | P2 | `--race` not on all test targets | ✅ **Fixed** — `--race` added to `test-unit-%` wildcard, `test-unit-shared-packages`, `test-unit-datastorage`, `test-unit-authwebhook`, `test-integration-shared`, `test-integration-datastorage` (E2E targets correctly excluded due to overhead) |
| P3-CON-1 | P3 | Thundering herd on `CheckExistence` | ✅ **Fixed** — `singleflight.Group` wraps the K8s API call, coalescing concurrent lookups for the same cache key |
| P2-ERR-1 | P2 | Batch validation error too generic | ✅ **Fixed** — Error field now includes `fmt.Sprintf("Signal validation failed: %s", valErr.Error())` |
| P2-SPEC-1 | P2 | HTTP 207 JSON not documented | ✅ **Fixed** — GoDoc on `BatchProcessingResponse` now notes: "This is a JSON-encoded body (Content-Type: application/json), not the RFC 4918 (WebDAV) XML multi-status format." |
| P2-API-1/2 | P2 | Test-only exports in production package | ✅ **Addressed** — Moved to `pkg/gateway/testing_exports.go`. Functions remain exported (required by external test package), but separated from `server.go`. Acceptable trade-off. |

### New Code in This Round

The team also added a **Service→workload traversal** feature (`8f5c9a2`) that resolves `Service` targets through `spec.selector → Pods → ownerRefs` — a meaningful improvement for Prometheus scrape-infrastructure labels (e.g., etcd-defrag scenario). The implementation:
- Uses `fallbackReader` (uncached) for the Service GET and Pod list — correct for fresh data
- Sorts pods deterministically by name — good for consistent fingerprinting
- Gracefully degrades when no fallbackReader, no selector, or no matching pods
- `ResolveFingerprint` now returns the resolved `ResourceIdentifier` so adapters can set the correct target kind

### Minor Observation (not blocking)

**P3-NEW-1 (nit):** `testing_exports.go` is not behind a build tag, so the `SetCacheReadyForTesting` / `SetShuttingDownForTesting` / `NewMinimalServerForReadinessTest` functions are compiled into production binaries. Since they don't allocate resources or have side effects, and Go's dead-code elimination handles unused functions, this is acceptable. A future cleanup could add `//go:build testing` if desired.

**P3-NEW-2 (nit):** The `isValidK8sName` regex enforces lowercase-only, which is correct for K8s resource names and namespaces. However, Prometheus label values like `StatefulSet` (used as resource identifiers in some recording rules) would fail validation. In `extractTargetResource`, this correctly skips non-RFC-1123 values, but it's worth adding a comment explaining the lowercase-only intent.

### Verdict: **approve-with-nits**

All P1 and P2 findings are resolved. CI is in progress (builds pass, unit tests pending). The P3 nits above are non-blocking. Ship it once CI is green.

Issue #74, opened 2026-02-11
feat(we): DD-EVENT-001 v1.1 -- Wire 5 K8s events in WorkflowExecution controller (BR-WE-095)


jordigilh:
## Summary

Wire 5 new K8s events in the WorkflowExecution controller per DD-EVENT-001 v1.1.

## Business Requirement

**BR-WE-095**: WorkflowExecution K8s Event Observability

## Events to Implement

| Event | Type | Priority | Emission Point |
|---|---|---|---|
| `CooldownActive` | Normal | P2 | reconcilePending: cooldown active, requeuing |
| `WorkflowValidated` | Normal | P2 | reconcilePending: spec validation passed |
| `WorkflowValidationFailed` | Warning | P2 | reconcilePending: spec validation failed (before MarkFailedWithReason) |
| `CleanupFailed` | Warning | P4 | ReconcileTerminal/ReconcileDelete: exec.Cleanup error |
| `PhaseTransition` | Normal | P3 | Intermediate phase changes (shared constant) |

## Existing Coverage (7 events, v1.0)

ExecutionCreated, LockReleased, WorkflowExecutionDeleted, PipelineRunCreated, WorkflowCompleted, WorkflowFailed (x2)

## Test Plan

- `docs/testing/DD-EVENT-001/TP-EVENT-WE.md`
- UT: FakeRecorder assertions (extends `test/unit/workflowexecution/controller_test.go`)
- IT: Business flow event trails

## Related

- Tracking: #71
- DD-EVENT-001 v1.1: `docs/architecture/decisions/DD-EVENT-001-controller-event-registry.md`

jordigilh:
Completed: 5 WE events implemented via TDD. Unit tests (7) + integration tests (4) all pass.

Issue #60, opened 2026-02-09
feat(notification): implement PagerDuty delivery channel


jordigilh:
## Summary

Implement PagerDuty as a delivery channel in the Notification service, enabling incident alerts to be routed to PagerDuty via the Events API v2.

## Context

The NT service delivery architecture (`pkg/notification/delivery/`) uses a pluggable `Service` interface pattern. PagerDuty already has **partial scaffolding** in place:

- `PagerDutyConfig` struct in `pkg/notification/routing/config.go` (service_key, routing_key, severity)
- Routing resolver recognizes `"pagerduty"` and maps to `ChannelWebhook`
- `Receiver.PagerDutyConfigs` field in routing config

**What's missing**: a `pagerduty.go` delivery service implementing the `Service` interface, and registration with the orchestrator.

### Delivery Service (`pkg/notification/delivery/pagerduty.go`)

- Implement `Service` interface (`Deliver(ctx, notification) error`)
- POST to PagerDuty Events API v2 (`events.pagerduty.com/v2/enqueue`)
- Map NotificationRequest fields to PagerDuty event payload (routing_key, event_action, severity, summary, source, component)
- Error classification following Slack pattern: 5xx/429 = retryable, 4xx = permanent
- Consider using official Go client `github.com/PagerDuty/go-pagerduty` or raw HTTP

### CRD/API Changes

- Evaluate whether to add a dedicated `ChannelPagerDuty` constant or continue routing through `ChannelWebhook`

### Registration

- Register with orchestrator via `RegisterChannel("pagerduty", pagerdutyService)` in controller setup

### Mock Server Options (triaged)

**Option A (recommended): Prism + official OpenAPI spec**
- PagerDuty publishes their full REST API v2 OpenAPI spec at `github.com/PagerDuty/api-schema`
- Use Stoplight Prism (`stoplight/prism` Docker image) to generate a spec-compliant mock with zero custom code
- `docker run --rm -p 4010:4010 stoplight/prism mock pagerduty-events-v2.yaml`
- Pros: Spec-compliant responses, zero maintenance, validates request format
- Cons: Requires Docker, heavier than a simple Go mock

**Option B: Custom Go mock server**
- Build a lightweight mock following `test/services/mock-llm/` pattern
- Simple HTTP handler that accepts Events API v2 payloads and returns `{ "status": "success", "dedup_key": "...", "message": "Event processed" }`
- Pros: Consistent with existing test infrastructure, lightweight
[...]

Issue #45, opened 2026-02-06
feat(workflowexecution): Ansible/AAP execution backend for enterprise remediation workflows


jordigilh:
## Summary

Add Ansible as a workflow execution backend, enabling remediation workflows written as Ansible playbooks to be executed by the WorkflowExecution controller. This builds on the execution backend abstraction introduced in #44 (K8s Job backend).

Three implementation approaches are proposed for WE team review. They are not mutually exclusive — Option A is a natural first step that Option B/C can complement later.

## Prerequisite

- #44 (K8s Job backend + Executor interface) must land first. All three options below depend on the `Executor` interface and `executionEngine` CRD field from that issue.

## Option A: Ansible via K8s Job (Minimal Effort)

Run a container image with `ansible-playbook` installed. The WE controller creates a K8s Job whose entrypoint executes the playbook.

```
WE Controller → creates batchv1.Job → Pod runs ansible-playbook → exits 0/1
```

### What's needed
- A convention for passing playbook parameters (env vars, mounted ConfigMap, or CLI args)
- A standard Ansible runner container image (e.g., based on `quay.io/ansible/ansible-runner`)
- Documentation on how to package playbooks into container images

### Characteristics
| Aspect | Assessment |
|--------|-----------|
| New controller code | **None** — reuses the Job executor from #44 |
| Cluster dependencies | **None** — K8s Jobs are native |
| Testability in Kind | **Full** — same as any Job |
| Execution visibility | **Low** — controller sees Job success/failure, not individual Ansible tasks |
| Enterprise value | **Medium** — works for simple playbooks, no AAP integration |

### Verdict
Effectively free once #44 lands. The Ansible-specific logic lives in the container image, not the controller. A workflow catalog entry would have `execution_engine: "job"` and `container_image: "your-ansible-runner:v1"`.

---

## Option B: AAP/AWX REST API Integration (Enterprise)

For organizations already running Red Hat Ansible Automation Platform (AAP) or AWX, the WE controller calls the AAP REST API to launch a job template.

```
WE Controller → HTTP POST to AAP/AWX /api/v2/job_templates/{id}/launch/
             → polls job status
             → maps AAP states to WFE phases
```

### What's needed
- New `AAPExecutor` implementing the `Executor` interface from #44
- AAP connection configuration (URL, auth token, CA cert) — likely via a Secret reference
- HTTP client for AAP REST API with retry/timeout handling
[...]

Issue #42, opened 2026-02-06
Batch audit write API returns event IDs in non-deterministic order


jordigilh:
## Summary

The `POST /api/v1/audit/events/batch` endpoint returns `event_ids` in non-deterministic order, breaking the positional correspondence between the request array and the response array.

**Affected component**: `pkg/datastorage/repository/audit_events_repository.go` — `CreateBatch()`

**Business Requirements**: DD-AUDIT-002 (StoreBatch interface), BR-AUDIT-001 (complete audit trail)

## Problem

The `CreateBatch` repository method groups events by `correlation_id` into a Go `map[string][]*AuditEvent` (for SOC2 Gap #9 hash chain integrity), then iterates the map to insert events and build the returned `createdEvents` slice:

```go
eventsByCorrelation := make(map[string][]*AuditEvent)
for _, event := range events {
    eventsByCorrelation[event.CorrelationID] = append(eventsByCorrelation[event.CorrelationID], event)
}
// ...
for correlationID, correlationEvents := range eventsByCorrelation {
    // inserts events, appends to createdEvents
}
```

Go maps have **non-deterministic iteration order**. The `createdEvents` returned to the handler is therefore in random order. The handler then builds `eventIDs[i] = createdEvents[i].EventID` and returns this to the client — breaking the contract that `eventIDs[0]` corresponds to `request[0]`.

## Impact

- Callers of the batch API cannot rely on the response order to correlate which `event_id` belongs to which submitted event.
- The E2E test `14_audit_batch_write_api_test.go` correctly validates this contract and fails intermittently depending on map iteration order (flake via `FlakeAttempts(3)`).

## Proposed Fix

Track each event's original position in the input slice. After all correlation-grouped inserts complete, reorder `createdEvents` to match the original input order before returning.

Approach: assign each event an index before grouping, then use that index to place the event in the correct position in the result slice.

```go
type indexedEvent struct {
    index int
    event *AuditEvent
}
// ... group indexedEvents by correlation_id ...
// ... after insert, place into result[ie.index] = ie.event ...
```

This preserves the hash chain grouping needed for SOC2 Gap #9 while restoring positional correspondence in the response.

## Acceptance Criteria

- [ ] `CreateBatch` returns events in the same order as the input slice
- [ ] SOC2 Gap #9 hash chain integrity is maintained (events with same correlation_id are still chained correctly)
[...]

PR #1057, opened 2026-05-07
Fix SA Token Refresh and Audit 401 Handling (#1055, #1056)

### LINKED ISSUE #1055: kubernaut-agent: SA token cached in memory, expires after 1h causing silent 401 failures
## Summary

The kubernaut-agent caches the projected service account token in memory at startup and never re-reads it from the projected volume after Kubernetes rotates it. Since the token has `expirationSeconds: 3600` (1 hour), all agent-to-data-storage API calls start returning **401 Unauthorized** after ~1 hour of uptime.

## Impact

- **Critical**: After 1 hour, the agent cannot call `list_available_actions`, `list_workflows`, or write audit events to data-storage-service. This causes every AI analysis to produce `ManualReviewRequired` / `NoMatchingWorkflows` even when a perfectly matching workflow exists — the LLM literally cannot see any workflows.
- The RCA phase (kubectl/prometheus tools) is unaffected since those don't go through data-storage.
- This silently affects **all scenarios** running after the first hour, not just a specific one.

## Evidence

From data-storage-service logs for agent pod IP `10.135.2.43`:

| Time (UTC) | Event |
|---|---|
| 14:27 | Pod started, token valid |
| 14:27–15:25 | All requests succeed (201/200) |
| ~15:27 | Token expires (expirationSeconds: 3600) |
| 16:22 | First 401 from agent |
| 17:55 | `list_available_actions` fails 3x with 401 during `etcd-defrag-forecast` workflow discovery |
| 17:55 | `list_workflows` (`GET /api/v1/workflows/actions/EtcdDefrag`) fails 3x with 401 |
| 17:56 | LLM gives up → `submit_result_no_workflow` → `NoMatchingWorkflows` |

Agent logs:
```
{"level":"error","tool":"list_available_actions","error":"listing action types: decode response: unexpected status code: 401"}
{"level":"error","tool":"list_workflows","error":"listing workflows: decode response: unexpected status code: 401"}
```

## Projected volume config

```yaml
volumes:
- name: sa-token
  projected:
    sources:
    - serviceAccountToken:
        audience: https://kubernetes.default.svc
        expirationSeconds: 3600
        path: token
```

## Expected behavior

The agent should periodically re-read the token from `/var/run/secrets/kubernetes.io/serviceaccount/token` (e.g., using a file watcher or reading on each request), so that kubelet-rotated tokens are picked up automatically.

## Workaround

Restart the kubernaut-agent pod to get a fresh token. This only buys another hour.

## Discovered during

Validation of `etcd-defrag-forecast` scenario on OCP (kubernaut-demo-scenarios).

### LINKED ISSUE #1056: kubernaut-agent: audit event batches silently dropped on non-retryable 401 errors
## Summary

When the kubernaut-agent's SA token expires (see #1055), the audit store classifies 401 errors as "non-retryable (invalid data)" and **silently drops** the entire audit batch without any user-visible indication. This means tool call events (`list_available_actions`, `list_workflows`, etc.) are never persisted, making it extremely difficult to diagnose workflow matching failures from audit traces alone.

## Impact

- **Observability gap**: When investigating `NoMatchingWorkflows` outcomes, the audit_events table shows no tool call events at all — making it appear as if the LLM never attempted workflow discovery, when in reality all calls failed with 401.
- Diagnosis required correlating kubernaut-agent container logs with data-storage request logs to find the actual root cause.

## Evidence

Agent logs show batch drops:
```
{"level":"error","msg":"Failed to write audit batch","batch_size":2,"error":"Data Storage Service returned status 401: decode response: unexpected status code: 401"}
{"level":"error","msg":"Dropping audit batch due to non-retryable error (invalid data)","batch_size":2,"is_4xx_error":true}
```

The batch is dropped because 401 is treated as a 4xx "invalid data" error. However, 401 is an **auth error**, not a data validation error — it should either:
1. Trigger token refresh and retry, or
2. At minimum, be logged as a distinct auth failure (not "invalid data")

## Suggested fix

- Distinguish 401/403 (auth errors) from 400/422 (data validation errors) in the retry logic
- For 401: attempt to re-read the SA token from disk and retry
- For persistent auth failures: emit a warning metric or condition on the pod status

## Related

- #1055 (root cause: SA token not refreshed from projected volume)

## Discovered during

Validation of `etcd-defrag-forecast` scenario on OCP (kubernaut-demo-scenarios).

jordigilh:
## Summary

- **#1055**: `bearerTransport` in `kubernaut-agent` read the SA token once at startup; after ~1h when kubelet rotates the projected token, all DataStorage and audit calls fail with 401 Unauthorized. Replaced with `AuthTransport` which re-reads from disk every 5 minutes and invalidates the cache immediately on 401 responses.
- **#1056**: All 4xx HTTP errors were classified as non-retryable ("invalid data"), including 401/403. Audit batches were silently dropped. Now 401/403 are classified as retryable auth errors with distinct logging for SRE diagnosis.

## Changes

| File | Change |
|------|--------|
| `pkg/shared/auth/transport.go` | Add `NewServiceAccountTransportWithPath` constructor, `invalidateTokenCache` method, 401 response handling in `RoundTrip`, `TokenInvalidationCount()` counter. Updated DD-AUTH-005 doc. |
| `pkg/audit/errors.go` | Add `IsAuthError()` method + package-level helper. Update `IsRetryable()` for 401/403. Document 401-vs-403 asymmetry. |
| `pkg/audit/store.go` | Auth-specific log in `writeBatchWithRetry`, clarified non-retryable drop message. |
| `cmd/kubernautagent/main.go` | Replace `bearerTransport` with `auth.NewServiceAccountTransportWithPath` in `initDSClients` + `buildAuditStore`. Remove `bearerTransport` type. Add `os.Stat` startup warning. |
| `test/unit/shared/auth/transport_test.go` | 9 new tests: custom path, 401 invalidation + counter, token rotation, non-401 cache, concurrent 401 storm, nil base, empty path, file deleted mid-operation. |
| `test/unit/audit/errors_test.go` | 9 updated/new tests: `IsRetryable` flips for 401/403, `IsAuthError` coverage, regression guards. |
| `test/unit/audit/store_test.go` | 3 new tests: 401 retry succeeds on 2nd attempt, auth-specific log message, persistent 401 exhaustion. |
| `docs/tests/1055-1056/TEST_PLAN.md` | IEEE 829 test plan (NEW). |

## Root Cause

```
kubelet rotates token → bearerTransport holds stale string → DS returns 401
→ audit store: "non-retryable 4xx" → DROP batch silently
→ LLM tools: fail → NoMatchingWorkflows
```

## Behavioral Change Note

**Audit retry latency on auth failures**: Previously, 401/403 batches were immediately dropped (~0s). Now they are retried up to MaxRetries (default 3) with exponential backoff (1s + 4s + 9s = ~14s worst case before drop). During a sustained auth outage, the background writer will spend ~14s per batch in retry backoff. This is acceptable (retrying is strictly better than silent drop) but operators should expect brief audit write latency spikes during token rotation windows.

## 401 vs 403 Design Asymmetry

- **401 (Unauthorized)**: `AuthTransport` invalidates the token cache → next retry re-reads from disk → self-heals.
- **403 (Forbidden)**: `AuthTransport` does NOT invalidate cache (refresh won't fix a permissions issue). Retries for 403 rely on RBAC propagation delays resolving between attempts. If persistent, batch is dropped after MaxRetries.

Both are classified as retryable in `IsAuthError()` to cover transient RBAC propagation delays, but only 401 triggers token refresh.

## Test plan

- [x] 21 new/updated unit tests covering all new behavior
- [x] All tests pass with `-race` (0 races)
- [x] `go build ./...` — 0 errors
- [x] `golangci-lint` — 0 new errors
- [x] `bearerTransport` — 0 references in production code
- [x] 100 Go Mistakes audit (#42, #52, #57, #58, #60, #74, #45) — all verified
- [x] 9-category checkpoint audit passed at all 3 checkpoints

Closes #1055
Closes #1056

### DIFF
--- a/cmd/kubernautagent/main.go
+++ b/cmd/kubernautagent/main.go
@@ -629,6 +629,11 @@ func initDSClients(cfg *kaconfig.Config, infra *k8sInfra, logger logr.Logger) *d
 	authTransport := auth.NewServiceAccountTransportWithPath(
 		cfg.Integrations.DataStorage.SATokenPath, dsBase)
 
+	if _, err := os.Stat(cfg.Integrations.DataStorage.SATokenPath); err != nil {
+		logger.Info("WARNING: SA token file not found at startup, DS API calls will fail auth until file appears",
+			"token_path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
+	}
+
 	var opts []ogenclient.ClientOption
 	opts = append(opts, ogenclient.WithClient(&http.Client{
 		Transport: authTransport,
@@ -747,6 +752,11 @@ func buildAuditStore(cfg *kaconfig.Config, logger logr.Logger) (audit.AuditStore
 
 	auditAuthTransport := auth.NewServiceAccountTransportWithPath(
 		cfg.Integrations.DataStorage.SATokenPath, auditBase)
+
+	if _, err := os.Stat(cfg.Integrations.DataStorage.SATokenPath); err != nil {
+		logger.Info("WARNING: SA token file not found at startup, audit batch writes will fail auth until file appears",
+			"token_path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
+	}
 	logger.Info("audit store auth configured (AuthTransport with token refresh)",
 		"token_path", cfg.Integrations.DataStorage.SATokenPath)
 
--- a/cmd/kubernautagent/main.go
+++ b/cmd/kubernautagent/main.go
@@ -632,6 +632,9 @@ func initDSClients(cfg *kaconfig.Config, infra *k8sInfra, logger logr.Logger) *d
 	if _, err := os.Stat(cfg.Integrations.DataStorage.SATokenPath); err != nil {
 		logger.Info("WARNING: SA token file not found at startup, DS API calls will fail auth until file appears",
 			"token_path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
+	} else {
+		logger.Info("DS client auth configured (AuthTransport with token refresh)",
+			"token_path", cfg.Integrations.DataStorage.SATokenPath)
 	}
 
 	var opts []ogenclient.ClientOption
@@ -639,8 +642,6 @@ func initDSClients(cfg *kaconfig.Config, infra *k8sInfra, logger logr.Logger) *d
 		Transport: authTransport,
 		Timeout:   defaultDSClientTimeout,
 	}))
-	logger.Info("DS client auth configured (AuthTransport with token refresh)",
-		"token_path", cfg.Integrations.DataStorage.SATokenPath)
 
 	ogenClient, err := ogenclient.NewClient(cfg.Integrations.DataStorage.URL, opts...)
 	if err != nil {
@@ -750,15 +751,21 @@ func buildAuditStore(cfg *kaconfig.Config, logger logr.Logger) (audit.AuditStore
 		return audit.NopAuditStore{}, nop
 	}
 
+	// PROD-R3-1: This creates a second AuthTransport with its own independent token cache.
+	// The DS ogen client (initDSClients) and audit store each have separate 5-min TTL caches.
+	// After token rotation, each invalidates independently on its own first 401. Audit writes
+	// may lag up to 5 min behind DS client recovery. Acceptable for v1; a shared cache is a
+	// future improvement.
 	auditAuthTransport := auth.NewServiceAccountTransportWithPath(
 		cfg.Integrations.DataStorage.SATokenPath, auditBase)
 
 	if _, err := os.Stat(cfg.Integrations.DataStorage.SATokenPath); err != nil {
 		logger.Info("WARNING: SA token file not found at startup, audit batch writes will fail auth until file appears",
 			"token_path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
+	} else {
+		logger.Info("audit store auth configured (AuthTransport with token refresh)",
+			"token_path", cfg.Integrations.DataStorage.SATokenPath)
 	}
-	logger.Info("audit store auth configured (AuthTransport with token refresh)",
-		"token_path", cfg.Integrations.DataStorage.SATokenPath)
 
 	dsClient, err := sharedaudit.NewOpenAPIClientAdapterWithTransport(
 		cfg.Integrations.DataStorage.URL, 5*time.Second, auditAuthTransport,
--- a/cmd/kubernautagent/main.go
+++ b/cmd/kubernautagent/main.go
@@ -623,22 +623,19 @@ func initDSClients(cfg *kaconfig.Config, infra *k8sInfra, logger logr.Logger) *d
 
 	const defaultDSClientTimeout = 30 * time.Second
 
+	// #1055: Use AuthTransport with configurable token path instead of static bearerTransport.
+	// AuthTransport re-reads the SA token from disk every 5 minutes (or immediately after
+	// a 401 response), so the agent survives kubelet token rotation.
+	authTransport := auth.NewServiceAccountTransportWithPath(
+		cfg.Integrations.DataStorage.SATokenPath, dsBase)
+
 	var opts []ogenclient.ClientOption
-	if tokenData, err := os.ReadFile(cfg.Integrations.DataStorage.SATokenPath); err == nil && len(tokenData) > 0 {
-		token := string(tokenData)
-		opts = append(opts, ogenclient.WithClient(&http.Client{
-			Transport: &bearerTransport{base: dsBase, token: token},
-			Timeout:   defaultDSClientTimeout,
-		}))
-		logger.Info("DS client auth configured (DD-AUTH-014)", "token_path", cfg.Integrations.DataStorage.SATokenPath)
-	} else {
-		opts = append(opts, ogenclient.WithClient(&http.Client{
-			Transport: dsBase,
-			Timeout:   defaultDSClientTimeout,
-		}))
-		logger.Info("SA token not available for DS client — DS API calls may fail auth",
-			"path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
-	}
+	opts = append(opts, ogenclient.WithClient(&http.Client{
+		Transport: authTransport,
+		Timeout:   defaultDSClientTimeout,
+	}))
+	logger.Info("DS client auth configured (AuthTransport with token refresh)",
+		"token_path", cfg.Integrations.DataStorage.SATokenPath)
 
 	ogenClient, err := ogenclient.NewClient(cfg.Integrations.DataStorage.URL, opts...)
 	if err != nil {
@@ -681,19 +678,6 @@ func buildDSBaseTransport(caFile string, cbCfg kaconfig.CircuitBreakerCfg) (http
 		FailureThreshold: cbCfg.FailureThreshold,
 		FailureRatio:     cbCfg.FailureRatio,
 	}), nil
-}
-
-// bearerTransport injects a Kubernetes ServiceAccount Bearer token into
-// every outbound HTTP request (DD-AUTH-014).
-type bearerTransport struct {
-	base  http.RoundTripper
-	token string
-}
-
-func (t *bearerTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	r := req.Clone(req.Context())
-	r.Header.Set("Authorization", "Bearer "+t.token)
-	return t.base.RoundTrip(r)
 }
 
 // buildEnricher creates the enrichment.Enricher when DS clients are available.
@@ -753,26 +737,21 @@ func buildAuditStore(cfg *kaconfig.Config, logger logr.Logger) (audit.AuditStore
 		return audit.NopAuditStore{}, nop
 	}
 
-	// Use the same auth transport as initDSClients: read SA token from the
-	// configured path and inject as Bearer header on every request.
+	// #1055: Use AuthTransport for audit store (same pattern as initDSClients).
+	// AuthTransport handles token refresh automatically via 5-minute cache + 401 invalidation.
 	auditBase, tlsErr := sharedtls.DefaultBaseTransport()
 	if tlsErr != nil {
 		logger.Error(tlsErr, "failed to create TLS-aware transport for audit store")
 		return audit.NopAuditStore{}, nop
 	}
 
-	var transport http.RoundTripper
-	if tokenData, err := os.ReadFile(cfg.Integrations.DataStorage.SATokenPath); err == nil && len(tokenData) > 0 {
-		transport = &bearerTransport{base: auditBase, token: string(tokenData)}
-		logger.Info("audit store auth configured (same SA token as DS client)",
-			"token_path", cfg.Integrations.DataStorage.SATokenPath)
-	} else {
-		logger.Info("SA token not available for audit store — batch writes may fail auth",
-			"path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
-	}
+	auditAuthTransport := auth.NewServiceAccountTransportWithPath(
+		cfg.Integrations.DataStorage.SATokenPath, auditBase)
+	logger.Info("audit store auth configured (AuthTransport with token refresh)",
+		"token_path", cfg.Integrations.DataStorage.SATokenPath)
 
 	dsClient, err := sharedaudit.NewOpenAPIClientAdapterWithTransport(
-		cfg.Integrations.DataStorage.URL, 5*time.Second, transport,
+		cfg.Integrations.DataStorage.URL, 5*time.Second, auditAuthTransport,
 	)
 	if err != nil {
 		logger.Error(err, "failed to create DS audit client, falling back to nop")
--- a/cmd/kubernautagent/main.go
+++ b/cmd/kubernautagent/main.go
@@ -180,8 +180,21 @@ func main() {
 	phaseTools := investigator.DefaultPhaseToolMap()
 
 	k8sInfra := initK8sInfra(logger)
-	ds := initDSClients(cfg, k8sInfra, logger)
-	auditStore, auditCleanup := buildAuditStore(cfg, logger)
+
+	// #1055: Create a single shared TokenSource for all DS-bound HTTP clients.
+	// Both the ogen DS client and the audit store share this cache, so a 401 on
+	// either side immediately invalidates the token for both.
+	dsTokenSource := auth.NewTokenSource(cfg.Integrations.DataStorage.SATokenPath)
+	if _, err := os.Stat(cfg.Integrations.DataStorage.SATokenPath); err != nil {
+		logger.Info("WARNING: SA token file not found at startup, DS/audit API calls will fail auth until file appears",
+			"token_path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
+	} else {
+		logger.Info("SA token source configured (shared cache with token refresh)",
+			"token_path", cfg.Integrations.DataStorage.SATokenPath)
+	}
+
+	ds := initDSClients(cfg, k8sInfra, dsTokenSource, logger)
+	auditStore, auditCleanup := buildAuditStore(cfg, dsTokenSource, logger)
 	reg := buildToolRegistry(cfg, logger, k8sInfra, ds)
 	enricher := buildEnricher(cfg, ds, k8sInfra, auditStore, logger)
 	sanitizer := buildSanitizationPipeline(cfg, logger)
@@ -600,7 +613,7 @@ type dsClients struct {
 // or default /var/run/secrets/kubernetes.io/serviceaccount/token), the ogen
 // client is configured with a Bearer token transport so that all DS API calls
 // (including ListWorkflows for the workflow validator) pass authentication.
-func initDSClients(cfg *kaconfig.Config, infra *k8sInfra, logger logr.Logger) *dsClients {
+func initDSClients(cfg *kaconfig.Config, infra *k8sInfra, dsTokenSource *auth.TokenSource, logger logr.Logger) *dsClients {
 	if cfg.Integrations.DataStorage.URL == "" {
 		logger.Info("DataStorage URL not configured, DS adapters disabled")
 		return nil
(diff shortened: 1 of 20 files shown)


pkg/shared/auth/transport.go
```diff
@@ -140,7 +159,14 @@ func (t *AuthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	// This allows services to start before ServiceAccount token is mounted
 	// Also allows local development without Kubernetes
 
-	return t.base.RoundTrip(reqClone)
+	resp, err := t.base.RoundTrip(reqClone)
+	if err == nil && resp.StatusCode == http.StatusUnauthorized {
```
jordigilh, line 233: **[QE-HIGH-1]** 401-only invalidation asymmetry needs documentation.

`RoundTrip` invalidates the token cache only on **401**, not on **403**. Meanwhile, `audit/errors.go` classifies both 401 and 403 as retryable auth errors. This means:

- On 401: transport invalidates cache → next retry re-reads token from disk → self-heals. ✅
- On 403: transport does **not** invalidate cache → retries use the same cached token → all retries fail if the issue is credential-related.

The design choice is correct (401 = expired credentials, refresh helps; 403 = insufficient permissions, refresh won't help). However, this asymmetry is not documented anywhere and could surprise future maintainers.

**Action**: Add a brief code comment on this line or in the type doc explaining why only 401 triggers invalidation (not 403).

jordigilh (commented):
## Multi-Persona Readiness Audit

Performed QE, Product, Product Security, UX, and SRE audits on this PR. Inline comments below are the actionable findings. Overall confidence: **90%**.

**Verdict**: PASS with minor actions.

| Persona | Verdict |
|---------|---------|
| QE | PASS with 1 action |
| Product | PASS |
| Product Security | PASS |
| UX | PASS with 1 action |
| SRE | PASS with follow-ups |

cmd/kubernautagent/main.go
```diff
@@ -623,22 +623,19 @@ func initDSClients(cfg *kaconfig.Config, infra *k8sInfra, logger logr.Logger) *d
...
-			Transport: dsBase,
-			Timeout:   defaultDSClientTimeout,
-		}))
-		logger.Info("SA token not available for DS client — DS API calls may fail auth",
-			"path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
-	}
+	opts = append(opts, ogenclient.WithClient(&http.Client{
+		Transport: authTransport,
+		Timeout:   defaultDSClientTimeout,
+	}))
```
jordigilh, line 644: **[UX-MEDIUM-1 / SRE-MEDIUM-2]** Startup token existence warning removed.

The old code checked if the token file was readable at startup and logged a clear warning:
```
SA token not available for DS client — DS API calls may fail auth
```

The new code **always** logs `"DS client auth configured (AuthTransport with token refresh)"` regardless of whether the token file actually exists. If the SA volume isn't mounted, operators see a misleading success message and won't discover the problem until the first 401 error (which could be minutes after startup).

**Action**: Consider adding a best-effort existence check at startup:
```go
if _, err := os.Stat(cfg.Integrations.DataStorage.SATokenPath); err != nil {
    logger.Info("WARNING: SA token file not found at startup, auth will fail until file appears",
        "token_path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
}
```

This doesn't change the AuthTransport behavior (it still degrades gracefully) but restores operator visibility. Same applies to `buildAuditStore` below.

pkg/audit/errors.go
```diff
@@ -59,12 +59,24 @@ func (e *HTTPError) Error() string {
...
+		return true
+	}
 	return e.StatusCode >= 500 && e.StatusCode < 600
 }
 
+// IsAuthError returns true for authentication/authorization errors (401, 403).
+// These are classified separately from data errors (400, 422) because auth errors
+// are transient when services use file-based SA token rotation (#1056).
+func (e *HTTPError) IsAuthError() bool {
+	return e.StatusCode == 401 || e.StatusCode == 403
```
jordigilh, line 83: **[SEC-MEDIUM-2]** 403 classified as auth error but transport won't refresh token for it.

`IsAuthError()` returns true for both 401 and 403, and `IsRetryable()` returns true for auth errors. This means the audit store will retry on 403. However, `AuthTransport.RoundTrip` only invalidates the token cache on 401, **not** 403.

Consequence: if a 403 occurs, the audit store retries 3 times with the **same** cached token, all fail, and the batch is dropped. The retries are wasted work.

This is arguably correct (403 = authorization denied, not credential expiry), but should be explicitly documented here to avoid future confusion. Consider adding a comment:

```go
// Note: 403 is retryable because RBAC propagation delays can cause transient
// authorization failures. However, AuthTransport only invalidates the token
// cache on 401 (credential expiry), not 403 -- retries for 403 use the same
// token and rely on the permission being granted between attempts.
```

pkg/audit/store.go
```diff
@@ -547,19 +547,27 @@ func (s *BufferedAuditStore) writeBatchWithRetry(batch []*ogenclient.AuditEventR
 		cancel()
 
 		if err != nil {
-			s.logger.Error(err, "Failed to write audit batch",
-				"attempt", attempt,
-				"batch_size", len(batch),
-			)
+			// #1056: Distinguish auth errors from data errors for observability
+			if IsAuthError(err) {
```
jordigilh, line 551: **[QE-MEDIUM-1]** Missing test: all retries exhausted with persistent 401.

UT-AS-1056-001 covers "401 then success on retry." There is no test for the case where 401 persists across all MaxRetries attempts (e.g., the token file on disk is also stale).

Expected behavior in this path: batch is dropped, \`failedBatchCount\` increments, and the \`"AUDIT DATA LOSS: Dropping batch after max retries"\` log is emitted.

**Action**: Add a test (e.g., UT-AS-1056-003) where the mock client returns 401 for all attempts, and verify:
- \`failedBatchCount\` increments
- The "AUDIT DATA LOSS" log appears
- The batch is dropped (not stuck in an infinite loop)

pkg/shared/auth/transport.go
```diff
@@ -140,7 +159,14 @@ func (t *AuthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		// #1055: Invalidate cached token on 401 so the next request re-reads
+		// from disk. This handles kubelet SA token rotation where the file
+		// has been updated but our cache still holds the stale token.
+		t.invalidateTokenCache()
```
jordigilh: **[SRE-MEDIUM-1]** No observability for token refresh events.

When \`invalidateTokenCache()\` fires and the next request re-reads a fresh token from disk, there is no log, metric, or counter emitted. SREs have no visibility into how often token refreshes occur in production.

The test plan acknowledges this as follow-up (SRE-M1, section 4.2). Consider at minimum a \`V(1)\` debug log here:

\`\`\`go
t.invalidateTokenCache()
// Consider adding: log.V(1).Info("token cache invalidated due to 401 response")
\`\`\`

Alternatively, track a counter (\`tokenRefreshCount\`) that can be exposed via metrics later.

test/unit/shared/auth/transport_test.go
```diff
@@ -213,6 +217,242 @@ var _ = Describe("AuthTransport", func() {
...
+			defer server.Close()
+
+			client := &http.Client{Transport: transport}
+			resp, err := client.Get(server.URL)
+			Expect(err).ToNot(HaveOccurred())
+			defer func() { _ = resp.Body.Close() }()
+			Expect(resp.Header.Get("X-Echo-Auth")).To(Equal("Bearer nil-base-token"))
+		})
+
+		It("should handle empty token path gracefully", func() {
```
jordigilh, line 479: **[QE-MEDIUM-2]** Missing test: token file deleted mid-operation.

Good coverage of the empty-path edge case. Consider adding a sibling test for the scenario where the token file **exists** initially, gets cached, then is **deleted** (simulating SA volume unmount or permissions change).

Expected behavior: after cache expiry (or 401 invalidation), \`getServiceAccountToken\` returns \`""\` and the request proceeds without auth header (graceful degradation).

This would validate the robustness of the cache-miss → file-read-fail → empty-string fallback path.

pkg/audit/store.go
```diff
@@ -547,19 +547,27 @@ func (s *BufferedAuditStore) writeBatchWithRetry(batch []*ogenclient.AuditEventR
...
+				s.logger.Error(err, "Auth error writing audit batch (token may be expired, will retry)",
```
jordigilh: **[PROD-MEDIUM-1]** Behavioral change: audit retry latency on auth failures.

Previously, 401/403 batches were immediately dropped (~0s). Now they are retried up to MaxRetries (default 3) with exponential backoff (1s + 4s + 9s = ~14s worst case before drop).

During a sustained auth outage (e.g., SA token rotation + stale file), the background writer will spend ~14s per batch in retry backoff. If multiple batches queue during this time, the delay compounds.

This is acceptable for the fix (retrying is strictly better than silent drop), but worth noting in the PR description as a behavioral change so operators know to expect brief audit write latency spikes during token rotation.

pkg/shared/auth/transport.go
jordigilh: Agreed — the asymmetry was a gap in documentation. Added a detailed comment in `RoundTrip` (commit 84beb67) explaining why only 401 triggers invalidation: 401 = credential expired (refresh helps), 403 = insufficient permissions (refresh won't help). Also documented the same rationale in the `IsAuthError` doc comment in `errors.go`.

cmd/kubernautagent/main.go
jordigilh: Good catch — restored the startup visibility. Added `os.Stat` checks in both `initDSClients` and `buildAuditStore` (commit 84beb67). If the token file is missing at startup, operators now see:
```
WARNING: SA token file not found at startup, DS API calls will fail auth until file appears
```
AuthTransport still degrades gracefully at runtime, but the startup log gives operators immediate visibility.

pkg/audit/errors.go
jordigilh: Added the suggested documentation (commit 84beb67). The `IsAuthError` doc comment now explicitly explains:
- 401: AuthTransport invalidates cache → next retry re-reads from disk → self-heals
- 403: AuthTransport does NOT invalidate cache → retries rely on RBAC propagation delays resolving

Also updated the PR description with a dedicated "401 vs 403 Design Asymmetry" section.

pkg/audit/store.go
jordigilh: Good point — updated the PR description with a "Behavioral Change Note" section documenting the latency impact: worst case ~14s (1s + 4s + 9s backoff) per batch during sustained auth outage, vs immediate drop before. This is strictly better than silent data loss.

test/unit/shared/auth/transport_test.go
jordigilh: Added the test (commit 443b476): token file exists → gets cached → file deleted (simulating SA volume unmount) → 401 invalidates cache → next request proceeds with empty auth header (graceful degradation). Validates the cache-miss → file-read-fail → empty-string fallback path.

pkg/shared/auth/transport.go
jordigilh: Added UT-AS-1056-003 (commit 443b476): mock returns 401 for all attempts, verifies `BatchCount() == 0` (batch dropped, not stuck) and the "Dropping"/"max retries" log appears for SRE diagnosis.

pkg/audit/store.go
jordigilh: Added an atomic `TokenInvalidationCount()` counter on `AuthTransport` (commit 84beb67) — lightweight, thread-safe, and ready to be wired into Prometheus metrics. UT-AT-1055-002 now asserts the counter increments from 0 to 1 after a 401 response. No logger needed in the transport layer since the audit store already logs auth errors with full context.

docs/tests/1055-1056/TEST_PLAN.md
```diff
@@ -0,0 +1,223 @@
...
+| UT-AS-1056-001 | 401 error retries and succeeds on 2nd attempt | Mock returns HTTPError{401} once, then succeeds | `AttemptCount() >= 2`, `BatchCount() == 1` | #1056 |
+| UT-AS-1056-002 | Auth error logging contains diagnostic context | Mock returns HTTPError{401}; capture error logs | Log contains "auth" or "token" context | #1056 |
+
+---
+
+## 7. Pass/Fail Criteria
+
+### 7.1 Pass Criteria
+
+- All 16 test cases pass
```
jordigilh: **[NIT-1]** Test plan says 16 tests but PR now has 18.

UT-AS-1056-003 (persistent 401 exhausts all retries) and the file-deletion edge case test were added after the plan was written. Consider updating section 7.1 to say "All 18 test cases pass" and adding the two new tests to section 6.3 for traceability.

Non-blocking -- the actual tests are authoritative.

pkg/shared/auth/transport.go
```diff
@@ -86,6 +90,10 @@ type AuthTransport struct {
 	tokenCache      string
 	tokenCacheTime  time.Time
 	tokenCacheMutex sync.RWMutex
+
+	// Observability: counts how many times the cache was invalidated due to 401.
+	// Exposed via TokenInvalidationCount() for metrics/testing (SRE-M1).
+	tokenInvalidationCount int64
```
jordigilh, line 100: **[NIT-2]** Mixed synchronization pattern on tokenInvalidationCount.

invalidateTokenCache() increments via atomic.AddInt64 while holding tokenCacheMutex.Lock(), but TokenInvalidationCount() reads via atomic.LoadInt64 without the mutex. This is safe (atomics are self-synchronizing), but the mixed pattern could confuse future readers.

Consider adding a brief comment on this field clarifying that reads use atomic (no mutex needed) while writes happen under the mutex but use atomic for lock-free reader access.

Non-blocking.

cmd/kubernautagent/main.go
```diff
@@ -623,23 +623,25 @@ func initDSClients(cfg *kaconfig.Config, infra *k8sInfra, logger logr.Logger) *d
...
+			"token_path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
 	}
 
+	var opts []ogenclient.ClientOption
+	opts = append(opts, ogenclient.WithClient(&http.Client{
+		Transport: authTransport,
+		Timeout:   defaultDSClientTimeout,
+	}))
+	logger.Info("DS client auth configured (AuthTransport with token refresh)",
+		"token_path", cfg.Integrations.DataStorage.SATokenPath)
```
jordigilh: **[UX-R3-1]** MEDIUM: Startup warning followed by unconditional success log.

When the token file is missing, operators see two consecutive lines:

```
WARNING: SA token file not found at startup, DS API calls will fail auth until file appears
DS client auth configured (AuthTransport with token refresh)
```

The success message after the warning is contradictory. Same pattern in buildAuditStore below.

Consider either:
(a) Making the success log conditional (else branch of the os.Stat check), or
(b) Including a qualifier: "DS client auth configured (AuthTransport with token refresh, token pending first file read)"

pkg/audit/store.go
jordigilh: **[UX-R3-2]** LOW: Auth error log message slightly misleading for 403.

This log says "token may be expired" for both 401 and 403. For 401 this is accurate. For 403 (insufficient permissions), the issue is authorization, not token expiry.

Consider: "Auth error writing audit batch (token expired or insufficient permissions, will retry)"

**[SRE-R3-1]** LOW: Auth error log missing max_retries field.

This log includes attempt and batch_size but not max_retries. For SRE triage, "attempt 2 of 3" is more actionable than just "attempt 2". Consider adding "max_retries", s.config.MaxRetries to the structured fields.

Both non-blocking.

cmd/kubernautagent/main.go
```diff
@@ -623,23 +623,25 @@ func initDSClients(cfg *kaconfig.Config, infra *k8sInfra, logger logr.Logger) *d
...
-		opts = append(opts, ogenclient.WithClient(&http.Client{
-			Transport: dsBase,
-			Timeout:   defaultDSClientTimeout,
-		}))
-		logger.Info("SA token not available for DS client — DS API calls may fail auth",
-			"path", cfg.Integrations.DataStorage.SATokenPath, "error", err)
+	// #1055: Use AuthTransport with configurable token path instead of static bearerTransport.
+	// AuthTransport re-reads the SA token from disk every 5 minutes (or immediately after
+	// a 401 response), so the agent survives kubelet token rotation.
+	authTransport := auth.NewServiceAccountTransportWithPath(
```
jordigilh: **[PROD-R3-1]** LOW: Two independent AuthTransport instances with independent caches.

initDSClients creates authTransport and buildAuditStore creates auditAuthTransport. Each has its own 5-minute cache. If DS returns 401 to the ogen client (LLM tools), only authTransport invalidates its cache. auditAuthTransport keeps its stale token until it also receives a 401 or its own TTL expires.

This means audit writes could lag up to 5 minutes behind DS client recovery after token rotation. Acceptable for v1, but worth noting in the design docs. A future improvement could share a single AuthTransport instance (or its cache) between both clients.

Non-blocking.

docs/tests/1055-1056/TEST_PLAN.md
jordigilh: Updated (commit ba00f19): section 5.2 TDD table now says UT-AT-1055-001..008, UT-AE-1056-001..009, UT-AS-1056-001..003. Section 6.1 adds UT-AT-1055-006/007/008 (nil base, empty path, file deletion). Section 6.3 adds UT-AS-1056-003 (persistent 401 exhaustion). Section 7.1 pass criteria updated to 21 tests.

pkg/shared/auth/transport.go
jordigilh: Added clarifying comment on the field (commit ba00f19): "Writes use atomic.AddInt64 (under tokenCacheMutex for cache consistency). Reads use atomic.LoadInt64 (no mutex needed — atomics are self-synchronizing)." This makes the intent explicit for future readers.

cmd/kubernautagent/main.go
jordigilh: Good catch — went with option (a): the success log is now in the `else` branch of the `os.Stat` check in both `initDSClients` and `buildAuditStore` (commit ba00f19). Operators only see the success message when the file actually exists.

pkg/audit/store.go
jordigilh: Both addressed (commit ba00f19):
- **[UX-R3-2]**: Log message updated to "Auth error writing audit batch (token expired or insufficient permissions, will retry)" — accurate for both 401 and 403.
- **[SRE-R3-1]**: Added `"max_retries", s.config.MaxRetries` to the structured fields so SREs see "attempt 2, max_retries 3" in logs.

cmd/kubernautagent/main.go
jordigilh: Acknowledged in code (commit ba00f19). Added a comment above `auditAuthTransport` in `buildAuditStore` explaining: independent 5-min TTL caches, audit writes may lag up to 5 min behind DS client recovery after rotation, acceptable for v1, shared cache is a future improvement. Thanks for flagging.

Issue #1582, opened 2026-07-06
KA: add AWS Bedrock provider support via Bedrock's OpenAI-compatible + native Anthropic endpoints


jordigilh:
## Summary

Add full, working AWS Bedrock provider support to KA, using AWS Bedrock's OpenAI-compatible API surfaces plus `anthropic-sdk-go`'s native Bedrock support for Claude, rather than a bespoke SigV4/Bedrock-native-envelope client. This preserves (and improves on) the Bedrock model coverage KA's config currently declares via `langchaingo` — but note that **Bedrock is not exercised in any current `kubernaut-operator` deployment sample**; this is dormant, config-level-only capability today.

Sub-task of #1578. Design captured in [DD-LLM-006](../blob/main/docs/architecture/decisions/DD-LLM-006-bedrock-dual-client-routing.md).

> **Update (2026-07-06)**: no live AWS Bedrock access is currently available to validate this spec. The routing model below is sourced from AWS's own documentation plus cross-checking third-party sources (see DD-LLM-006), captured now so the requirement is clean, with **implementation deferred to its own PR** — decoupled from #1580/#1581's landing — once Bedrock test access is available. The original two-way (Claude / everything-else) routing model was revised to three-way after the doc spike surfaced a second OpenAI-compatible endpoint (see below); this is more accurate but still needs live-access verification per the checklist at the bottom.

## Why this is now cheap (evidence, revised after documentation spike)

AWS Bedrock exposes **two distinct endpoints** relevant here, per `docs.aws.amazon.com/bedrock/latest/userguide/{apis,inference-chat-completions-mantle,models-endpoint-availability}.html`:

| Endpoint | Base URL | Auth | Model coverage |
|---|---|---|---|
| `bedrock-runtime` (legacy/general — existing native integration path) | `https://bedrock-runtime.{region}.amazonaws.com/v1/chat/completions` (OpenAI-compatible) or native `Converse`/`InvokeModel` | AWS SigV4 or Bedrock API key (Bearer) | Models Bedrock has natively integrated: Claude (all variants), Nova, Llama, Titan, Cohere, most of Mistral's lineup, Qwen. AWS's own docs show `model: "anthropic.claude-sonnet-4-6"` used directly against this endpoint's Chat Completions path. |
[...]

Issue #1581, opened 2026-07-06
KA: replace langchaingo with a hand-rolled OpenAI-compatible client (shared with AF's adapter pattern)


jordigilh:
## Summary

KA should stop depending on `langchaingo` for OpenAI-protocol providers (native `openai`, self-hosted OpenAI-compatible endpoints, `azure`) and instead adopt the same architectural pattern AF already uses successfully in production: a hand-rolled, plain-`net/http` OpenAI-compatible client, with no third-party LLM framework dependency.

This is a sub-task of #1578 (reasoning/thinking token support) and a sibling of #1580 (langchaingo removal for the Anthropic-family path). Reasoning/round-trip support is the primary driver, but this also removes KA's dependency on an increasingly unmaintained library (see #1580 for the maintenance-health evidence).

## Existing prior art: AF's adapter

AF already has exactly this, in production, covering OpenAI, LlamaStack, vLLM, and Ollama with zero langchaingo dependency:

```1:7:pkg/apifrontend/launcher/openai/adapter.go
// Package openai provides an in-house adapter implementing the ADK model.LLM
// interface for OpenAI-compatible API endpoints (OpenAI, LlamaStack, vLLM, Ollama).
//
// The adapter uses net/http directly and accepts a custom *http.Client for
// transport chain injection (TLS CA, OAuth2, custom headers, circuit breaker)
// per Issue #1342.
```

It already handles: non-streaming + streaming (SSE) chat completions, tool-call accumulation across streamed chunks, JSON schema conversion for tool parameters, and the same transport-chain injection pattern (`WithHTTPClient`) KA's `llm_builder.go` already uses for TLS/OAuth2/circuit-breaker.

**What it does NOT have**: any reasoning/thinking field support (`reasoning`/`reasoning_content`) — request or response side. That's net-new work regardless of which codebase it lands in.

## Design question to resolve: shared library or parallel port?

AF's adapter implements ADK's `model.LLM` interface and speaks `genai.Content`/`genai.Part` (AF's native wire format). KA's `llm.Client` interface uses KA's own types (`llm.Message`, `llm.ChatRequest`/`llm.ChatResponse`) by design, to keep business logic isolated from any specific LLM framework (DD-HAPI-019). The two interfaces are not directly interchangeable.

[...]

Issue #1580, opened 2026-07-06
KA: remove langchaingo dependency for Anthropic-family providers (generalized native/Vertex/Bedrock client)


jordigilh:
## Summary

Remove `github.com/tmc/langchaingo` from KA (`kubernautagent`) entirely, replacing its role for the Anthropic-family provider paths (`provider: anthropic` native, `provider: vertex_ai`) with a single generalized custom client built on `anthropic-sdk-go` (already vendored).

> **Update (2026-07-06)**: spiked whether reusing `adk-anthropic-go/converters` for model-aware thinking mapping (as originally proposed below) would couple KA to the ADK agent framework. Initial read of `converters/response.go`'s `google.golang.org/adk/model` import suggested it would (Go compiles whole packages). Measuring the *actual* incremental cost to KA's own binary (`cmd/kubernautagent` is a separate deployable from AF's `cmd/apifrontend`) told a different story: **~89 KB binary size delta, and only one thin ADK types package (`adk/model`)** — none of the agent/runner/session/A2A-server code AF actually uses ADK for. Given that measured cost is negligible and avoids a real drift risk (two independently-maintained tier-detection tables as Anthropic ships new adaptive-capable models), **the original plan (reuse) is confirmed and locked in** — scope item 2 below is unchanged from the initial proposal. See [DD-LLM-005](../blob/main/docs/architecture/decisions/DD-LLM-005-model-aware-reasoning-support.md) for full before/after analysis. Business requirement: [BR-AI-086](../blob/main/docs/requirements/BR-AI-086-llm-reasoning-token-support.md) (not `BR-HAPI-263`, which is deprecated-HAPI-scoped and does not apply to KA).

This is a sub-task of #1578 (reasoning/thinking token support), split out because it's a distinct architectural change (dependency removal) with its own risk profile, not just a reasoning-specific fix.

## Why remove langchaingo (evidence)

- **Maintenance health**: `tmc/langchaingo`'s default branch has not been pushed to since **2026-01-11** (~6 months as of this writing), despite 404 open issues and 111 open PRs. Contributor concentration is heavy (`tmc`: 504 contributions vs. 188 for the #2 contributor).
[...]

PR #1598, opened 2026-07-06
feat(llm): reasoning-token support, langchaingo removal, shared openaicompat client (#1578, #1580, #1581)

### LINKED ISSUE #1578: KA: reasoning/thinking tokens not requested or preserved across LLM providers (model-aware config needed)
## Summary

KA's LLM clients never request, capture, or round-trip reasoning/thinking tokens for any provider. This issue is the parent/tracking issue for the problem statement, spike findings, and business rationale. **Implementation is broken into three sub-issues**, each an independent architectural piece that collectively also removes KA's dependency on `langchaingo` (see spike findings below for why that's a second, independently-motivated goal, not scope creep):

| Issue | Scope |
|---|---|
| **#1580** | Generalize `vertexanthropic.Client` (renamed `anthropicfamily`) into a single Anthropic-family client (native `anthropic` + `vertex_ai` + Bedrock-Claude auth modes), with model-aware thinking + signature round-trip **reusing `adk-anthropic-go/converters.ThinkingConfigToAnthropic` directly** (measured ~89 KB binary cost, no agent-framework code — see spike finding below). Removes langchaingo for this path. |
| **#1581** | Replace langchaingo with a hand-rolled OpenAI-compatible client (native `openai`, `azure`, self-hosted vLLM/Ollama/LlamaStack, DeepSeek), modeled on AF's existing production adapter — with reasoning-field support and round-trip-mode awareness (`none`/`plain`/`deepseek-conditional`) added, which AF's version doesn't have. Evaluates extracting a shared library between AF and KA. |
| **#1582** | AWS Bedrock provider support (dormant today, no active deployment), via #1580's client for Claude-on-Bedrock and #1581's client for everything else on Bedrock, using Bedrock's newer OpenAI-compatible endpoint. |

Current plan is to implement and land all three in the same PR for efficiency, given how tightly coupled they are (see dependency graph in #1582).

## Original gap (motivation)

- **Gap 1** — the request never asks for reasoning/thinking output, on any path.
- **Gap 2** — even where a provider/model returns reasoning content anyway, there's no code path to capture it or round-trip it across turns/phases (self-correction retries, Phase 1 RCA → Phase 3 Workflow Discovery).

### Original evidence: `vertex_ai` path (`vertexanthropic.Client`)

Nothing at all is wired up. `buildParams()` never sets `Thinking`, despite the vendored `anthropic-sdk-go v1.56.0`'s `MessageNewParams` having the field:

```201:212:pkg/kubernautagent/llm/vertexanthropic/client.go
func (c *Client) buildParams(req llm.ChatRequest) anthropic.MessageNewParams {
	params := anthropic.MessageNewParams{
		Model:     anthropic.Model(c.model),
		MaxTokens: int64(4096),
	}
	...
```

`mapResponse()` only handles `"text"` and `"tool_use"` blocks (no `"thinking"`/`"redacted_thinking"` case), and `convertAssistantMessage()` has no path to replay a `ThinkingBlockParam` (with its required `Signature`) back to Anthropic on the next turn.

### Prior-art risk

This is the same failure class kubernaut already hit once on the Anthropic side: issue **#1299** (`historySanitizer`), where orphaned `tool_use` blocks without a matching `tool_result` caused Anthropic 400s on replay. Anthropic has the equivalent hard requirement for thinking blocks preceding tool use — enabling `Thinking` without also fixing the round-trip would very likely reproduce that failure class in the self-correction loop, which currently has zero test coverage for this path.

## Spike findings summary (full detail in #1580/#1581/#1582)

> **Update (2026-07-06)**: `BR-HAPI-263` reference removed throughout this issue — it describes the deprecated Python HolmesGPT-API (HAPI) service's API surface, not KA. HAPI is no longer a service in Kubernaut and must not be referenced in current/future work. Replaced by a fresh requirement: [`BR-AI-086`](../blob/main/docs/requirements/BR-AI-086-llm-reasoning-token-support.md), backed by [`DD-LLM-004`](../blob/main/docs/architecture/decisions/DD-LLM-004-langchaingo-removal-generalized-clients.md), [`DD-LLM-005`](../blob/main/docs/architecture/decisions/DD-LLM-005-model-aware-reasoning-support.md), [`DD-LLM-006`](../blob/main/docs/architecture/decisions/DD-LLM-006-bedrock-dual-client-routing.md).
>
> **Update 2 (2026-07-06)**: closed out the spike deferred in requirement 7 below on `adk-anthropic-go/converters`, and it reversed the initial conclusion. Reading `converters/response.go`'s `google.golang.org/adk/model` import initially suggested importing the package would transitively adopt the ADK agent framework. Measuring the *actual* cost (building `cmd/kubernautagent` with/without the import, diffing package closures) showed something much smaller: **~89 KB binary size delta and one thin ADK types package (`adk/model`)** — none of the agent/runner/session/A2A-server code AF actually depends on ADK for. `google.golang.org/adk` is also already a module dependency (via AF), just not previously reachable from KA's own binary. Given this negligible measured cost vs. the real risk of two independently-drifting tier-detection tables, **#1580 now reuses `adk-anthropic-go/converters.ThinkingConfigToAnthropic` directly**, confined to the `anthropicfamily` adapter package (DD-HAPI-019's business-logic isolation rule is unaffected either way, since it only restricts `investigator.go`/`tools/`/`result/`, not the adapter layer).

1. **langchaingo's Anthropic backend cannot round-trip thinking** even though it reads it correctly — no `ThinkingContent`/signature part type exists in `llms.MessageContent`. Confirmed via source.
2. **langchaingo itself is a maintenance risk independent of reasoning**: default branch unpushed since 2026-01-11 (~6 months), 404 open issues, 111 open PRs, thin bus factor. Two independent unmerged PRs ([#1500](https://github.com/tmc/langchaingo/pull/1500), [#1505](https://github.com/tmc/langchaingo/pull/1505)) fix a `handleAIMessage` bug that drops content parts on replay — tested against **our exact production models** (`claude-sonnet-4-6`, `claude-opus-4-7`) by external contributors, unmerged, zero review activity.
3. **AF already solved the "no langchaingo" problem** for OpenAI-protocol providers: `pkg/apifrontend/launcher/openai/adapter.go` is a proven, production, hand-rolled `net/http` client covering OpenAI/LlamaStack/vLLM/Ollama with no reasoning support yet — the gap we'd be filling regardless of which codebase it lives in.
4. **Native OpenAI reasoning (o-series/GPT-5.x) requires the Responses API** (`encrypted_content`, order-sensitive, not portable across model swaps) — langchaingo only implements Chat Completions, which OpenAI's own docs confirm never round-trips reasoning. This is explicitly **out of scope** for #1581 (tracked as a future decision, would need `github.com/openai/openai-go` as a new dependency).
5. **AWS Bedrock now has its own OpenAI-compatible endpoint** (Bearer-token auth, no SigV4 needed) plus native Anthropic-on-Bedrock SDK support already in `anthropic-sdk-go` — meaning full Bedrock coverage (#1582) costs no new protocol code once #1580/#1581 exist.
6. **Capability auto-detection can't be the only mechanism** for self-hosted/custom models (langchaingo's `IsReasoningModel()`, `adk-anthropic-go`'s `supportsAdaptiveThinking()` are both name-pattern-based) — but this narrows to only being needed for #1581's self-hosted path; Anthropic-family (vendor-enum-driven) and native OpenAI (protocol-driven) don't need an override.

## Requirements for the fix (cross-cutting, apply across #1580/#1581/#1582)

1. **Provider-agnostic reasoning fields** in `llm.ChatOptions`/`llm.Message` (`pkg/kubernautagent/llm/types.go`) — business logic (`investigator`) only ever sees generic fields. Schema must accommodate: optional opaque signature/encrypted-content payload, plain text with no signature, and a round-trip-mode hint.
2. **Capability resolution with an explicit override for self-hosted models only** (#1581's scope) — auto-detection default, explicit CRD-backed config override. Requires a small `kubernaut-operator` CRD change (`LLMSpec`/`LLMPhaseOverrideSpec` in `api/v1alpha1/kubernaut_types.go`) — **to be filed as a separate, cross-linked issue in `kubernaut-operator` once the field shape is finalized during planning.**
3. Unsupported/unknown models must fail safely (skip the parameter) rather than error.
4. **Model-switch safety**: reasoning/encrypted-content state must be stripped, not replayed, across a phase handoff or hot-reload swap that changes the underlying model (`llm.SwappableClient`, `RuntimeParams`, `PhaseModels`).
5. Self-correction-loop + phase-handoff test coverage exercising thinking-enabled responses (currently zero coverage) before enabling by default, across all provider clients, to protect against the #1299-class regression.
6. Surface captured reasoning in the investigation audit trail (SOC2 CC7.2, BR-AUDIT-005).
7. Verify AF's actual behavior in a live session (Anthropic/Vertex and other active provider paths) rather than assuming correctness from library semantics alone.

## Expected improvement

Primary expected benefit is **investigation quality**, not auditability (auditability is real but secondary):

- Deeper multi-hypothesis reasoning when correlating tool-call evidence, historical context, and ambiguous root causes before committing to an RCA conclusion.
- Potentially more reliable self-correction: with visible reasoning, the model is more likely to genuinely reconsider *why* a validation failed rather than pattern-matching a superficial fix.
- Workflow-selection quality (Phase 3, currently `claude-haiku-4-5`) likely sees less benefit without an explicit manual thinking budget, since Haiku 4.5 isn't in the adaptive-capable tier.

**Caveat**: this is evidence-based inference from documented model behavior, not measured against kubernaut's own RCA outcomes yet. Recommend a time-boxed measurement spike (RCA accuracy/confidence with vs. without thinking) before enabling broadly.

## References

- Sub-issues: #1580, #1581, #1582
- BR-AI-086 (`docs/requirements/BR-AI-086-llm-reasoning-token-support.md`) — reasoning chain preservation across retries/phases, written fresh against KA's Go implementation (supersedes an earlier, incorrect `BR-HAPI-263` reference — that BR describes the deprecated Python HAPI service and does not apply to KA).
- Issue #1299 — prior, already-fixed instance of the same failure class.
- `config/samples/v1alpha1_kubernaut.yaml` (kubernaut-operator) — confirms production model configuration (`claude-sonnet-4-6` for RCA, `claude-haiku-4-5` for workflow discovery).
- `openai/codex#17541` — confirms cross-model `encrypted_content` decryption failure class in a comparable production agent.

jordigilh:
## Summary

Implements model-aware LLM reasoning/extended-thinking support across Kubernaut Agent (KA), fully deprecates the unmaintained `langchaingo` dependency, and extracts a shared OpenAI-compatible client (`pkg/shared/llm/openaicompat`) now used by both AI Frontend (AF) and KA. Originated from triage of [OLS-3442](https://redhat.atlassian.net/browse/OLS-3442) (reasoning tokens dropped at cache boundary in a peer project).

Closes #1578, #1580, #1581.

- **#1580** — Native Anthropic API-key auth + model-aware thinking (adaptive budget vs. manual budget, based on `adk-anthropic-go/converters`), replacing the Vertex-only `vertexanthropic` client (renamed `anthropicfamily`).
- **#1581** — Shared `openaicompat` package (wire types, HTTP, SSE streaming, tool-call accumulation, reasoning-mode detection incl. DeepSeek conditional replay) extracted from AF's adapter; both AF and KA now delegate to it via thin wrappers, closing the compatibility gap for local/self-hosted OpenAI-protocol models (Ollama, vLLM, LlamaStack, etc.).
- **#1578** — `langchaingo` fully removed (unmaintained, missing reasoning support); reasoning content is captured provider-agnostically into `llm.Message.Reasoning` / `InvestigationResult.Reasoning`, and surfaced in the audit trail (both per-turn `aiagent.llm.response` events and the `aiagent.rca.complete` summary) for SOC2 CC8.1 correlation-id reconstruction, with defensive truncation (40k chars) to bound audit payload size. Only human-readable reasoning text + a `redacted` flag are persisted — no opaque provider signature bytes.

Bedrock support (native + OpenAI-compatible routing) was scoped out to a follow-up issue per BR-AI-086; see `docs/architecture/decisions/DD-LLM-004/005/006.md` and `docs/requirements/BR-AI-086.md` for the full design record and follow-up issue list (CRD config, RCA-effectiveness measurement, demo-scenarios validation, additional SDK-gap issues).

## ⚠️ Provider support changes (breaking for some Helm deployments)

Removing `langchaingo` replaces its permissive provider dispatch with an explicit allowlist in `cmd/kubernautagent/llm_builder.go`. This was flagged by the Helm chart team ([review](https://github.com/jordigilh/kubernaut/pull/1598#issuecomment-4898816087)) as a real regression risk for existing deployments — thank you for catching this. Status, verified against code, provider by provider:

| Provider | Status after this PR | Action needed |
|---|---|---|
| `openai`, `anthropic`, `vertex_ai`, `openai_compatible` | ✅ Supported, unchanged | None |
| `mistral`, `ollama`, `huggingface`, local vLLM/LlamaStack | ✅ Supported, **config rename required** | Change `provider: mistral` (etc.) → `provider: openai_compatible` + set `endpoint` to the provider's `/v1` base URL. These all speak the same OpenAI Chat Completions wire protocol already handled by `openaicompat` — no code gap, just a config key change. |
| `bedrock` | ❌ Not supported (tracked, deferred) | Do not upgrade if using Bedrock today. Tracked in #1582 (dormant/config-only capability — not exercised by any current deployment sample per that issue). |
| `azure` | ❌ Not supported (tracked, deferred) | Do not upgrade if using Azure OpenAI today. Tracked in #1600 — Azure's URL shape (`/openai/deployments/{id}/...?api-version=...`) and `api-key` auth header are genuinely unhandled by `openaicompat` today (this is a real gap, not a config rename like mistral/ollama/etc). Filed as a shared fix for both AF and KA. |

Attempting to run with an unsupported `provider` value now fails fast at startup (`os.Exit(1)`) rather than silently misbehaving. Chart/schema/docs updates to reflect this table (enum constraint on `provider`, removal of the now-orphaned `azureApiVersion`/`bedrockRegion` static fields, corrected provider docs) are included in this PR — see the diff under `charts/kubernaut/` and `docs/services/kubernaut-agent/configuration-reference.md`.

## Milestones (TDD, RED→GREEN→REFACTOR per milestone)

1. **M1**: provider-agnostic reasoning fields on `llm.Message`/`ChatOptions`/`LLMConfig`
2. **M2**: `anthropicfamily` (renamed from `vertexanthropic`) — native API-key auth + model-aware thinking
3. **M3**: shared `openaicompat` client extracted from AF; AF + KA thin wrappers wired in `cmd/kubernautagent/llm_builder.go`
4. **M4**: `langchaingo` fully removed (client, adapter, tests, dependency)
5. **M5**: reasoning surfaced in the audit trail (per-turn + RCA-complete), with Phase 1 → Phase 3 handoff propagation and truncation/observability safeguards
6. **M6**: `E2E-KA-AUDIT-001` — mock-llm reasoning-capture scenario + E2E proof of correlation-id-based audit reconstruction (SOC2 CC8.1)

## Test plan

- [x] `go vet ./...` — clean
- [x] `golangci-lint run ./...` — 0 new issues (14 pre-existing issues remain, all in `docs/spikes/` and `scripts/validation/`, untouched by this PR)
- [x] `make test-unit-kubernautagent` — 31 suites, 101 specs, 0 failed (85.5% composite coverage)
- [x] `make test-unit-apifrontend` — 24 suites, 0 failed (83.4% composite coverage)
- [x] `make test-unit-datastorage` — 16 suites, 0 failed (58.8% composite coverage)
- [x] `go test ./pkg/shared/llm/... ./pkg/apifrontend/launcher/openai/...` — pass
- [x] `E2E-KA-AUDIT-001` — executed in CI; passes (fixed a bug found along the way — `sameKindValidationGate`/`apiVersionValidationGate` retry paths were dropping reasoning from the retry turn's response; see commit history)

## Notes for reviewers

- No regressions expected for existing Vertex/Anthropic or OpenAI-compatible deployments — `anthropicfamily` and `kaopenai` preserve prior wire behavior when `reasoning` is unset/disabled.
- `reasoning.enabled` is opt-in per `LLMConfig`/`LLMPhaseOverrideSpec` (hot-reloadable); unset models fall back to no-thinking behavior identical to pre-PR.
- Follow-up issues (filed separately, referenced in BR-AI-086): #1582 (Bedrock support), #1600 (Azure OpenAI support, AF+KA), #1599 (restart-required restriction for LLM provider/model hot-reload — a separate architectural inconsistency found post-merge-prep, unrelated to the provider-allowlist change above), `kubernaut-operator` CRD field additions, `kubernaut-demo-scenarios` validation scenario, RCA-effectiveness measurement, and remaining LLM-provider-SDK gaps (prompt caching, structured outputs, tool_choice forcing, token-counting API).


jordigilh:
## Chart-impact triage: this PR removes LLM providers the Helm chart currently documents as supported

Triaged this before merge since it changes `cmd/kubernautagent/llm_builder.go`'s provider dispatch. Findings, in case they weren't already on your radar for `charts/kubernaut/`:

### Breaking: KA's supported provider set shrinks, with no back-compat shim

`buildLLMClientFromConfig`'s switch narrows from "anything falls through to `langchaingo.New(cfg.Provider, ...)`" (which handled `azure`, `bedrock`, `mistral`, `ollama`, `huggingface`, `local`, etc.) to an explicit allowlist of `{vertex_ai, anthropic, openai, openai_compatible}`, with everything else now hitting:

```go
default:
    return nil, fmt.Errorf("unsupported LLM provider: %q", cfg.Provider)
```

I checked `mergeLLMConfig` and the rest of `cmd/kubernautagent` — there's no provider-string normalization/migration shim, so `provider: bedrock` (etc.) isn't silently remapped to a new equivalent value; it just errors.

That error path calls `os.Exit(1)` at startup (`cmd/kubernautagent/bootstrap.go:154-156`), so this is a hard crash-loop, not a soft degrade.

### Why this hits the chart specifically

`charts/kubernaut/values.yaml` (line 524) and `values.schema.json` (line 867) currently document/permit exactly the values being dropped:

```yaml
# Supported providers: openai, anthropic, bedrock, mistral, vertex_ai, local.
provider: ""
```

`values.schema.json`'s `provider` field has no `enum` constraint, so `helm install`/`upgrade` will accept `provider: bedrock` (etc.) without complaint — the chart-installed KubernautAgent pod will only fail after rollout, as a `CrashLoopBackOff`. Any existing Helm deployment on `bedrock`/`mistral`/`local`/`azure` will break on upgrade to a chart built on top of this PR.

Also flagging: `azureApiVersion` and `bedrockRegion` remain in `LLMConfig`/the chart schema (presumably kept for the Bedrock follow-up), but nothing in the new dispatch path consumes them anymore — they're orphaned config as of this PR. Separately (not introduced by this PR, but worth noting while here): neither the old AF adapter nor the new shared `openaicompat` client ever injects Azure's mandatory `?api-version=` query param, so `azureApiVersion` hasn't actually been wired end-to-end for a while.

Smaller/non-blocking: `docs/services/kubernaut-agent/configuration-reference.md`'s provider table (touched by this PR, line 177-180) still lists `bedrock`, `huggingface`, `vertex` as valid — looks like it wasn't fully updated for this change either.

### Suggested handling

Doesn't need to block this PR (it's `pkg`/`cmd` only, chart is untouched), but I'd like to land a fast-follow in `charts/kubernaut/` right after merge to:
1. Drop `bedrock`/`mistral`/`local` from the `provider` comment/description in `values.yaml`/`values.schema.json` (keep `azure` un-mentioned as a provider value too, since it's not in the new allowlist).
2. Either drop `azureApiVersion`/`bedrockRegion` from the chart (dead config) or add a code comment noting they're reserved for a future Bedrock follow-up.
3. Optionally expose the new `reasoning.{enabled,budgetTokens,capabilityOverride}` field for chart parity (non-urgent, safe default).

Let me know if you'd rather fold the chart fix into this PR instead — happy to open it either way.

jordigilh:
Thanks for the thorough catch — verified all of it against the code and folded a fix into this PR (610c30b9c):

- **`values.schema.json`**: `provider` (both `kubernautAgent.llm` and `alignmentCheck.llm`) is now enum-constrained to the actually-supported set (`openai`, `anthropic`, `vertex_ai`, `openai_compatible`). `helm template`/`install`/`upgrade` with `provider: bedrock` now fails immediately with a schema error instead of `CrashLoopBackOff` post-rollout — confirmed via `helm template ... --set kubernautAgent.llm.provider=bedrock`.
- **`values.yaml` / `configuration-reference.md`**: corrected to document, provider by provider:
  - `mistral`, `ollama`, `huggingface`, local vLLM/LlamaStack — **no code gap**, reachable via `provider: openai_compatible` + `endpoint` (they all speak the same OpenAI Chat Completions wire protocol `openaicompat` already handles). This is a config rename, not a functionality loss.
  - `bedrock` — genuinely not supported yet, tracked in #1582 (already existed, confirmed dormant/not exercised by any current deployment sample).
  - `azure` — genuinely not supported yet. Filed **new issue #1600** for this (shared fix needed in `pkg/shared/llm/openaicompat`, affects both AF and KA) since Azure needs a different URL shape + `api-version` query param + `api-key` auth header that the generic OpenAI-compatible client doesn't handle — this one's a real gap, not just a rename.
- **`azureApiVersion`/`bedrockRegion`**: left in place (not deleted) but doc'd as reserved/not-currently-consumed, since they're the intended re-entry point once #1600/#1582 land.

PR description updated with the full provider support matrix. Let me know if you'd like anything further before this merges.

PR #1601, opened 2026-07-07
fix(kubernautagent): wire ai.llm.reasoning config into real Anthropic requests

jordigilh:
## Summary

Post-merge follow-up to #1598. While auditing test coverage for the reasoning-token feature (BR-AI-086), found that `ai.llm.reasoning.enabled` / `budgetTokens` was **never actually wired into any real LLM call**:

- Every `llm.ChatOptions{}` construction site in `internal/kubernautagent/investigator/*.go` and `internal/kubernautagent/alignment/*.go` only sets `JSONMode`/`OutputSchema`/`MaxTokens`/`Temperature` — none ever set `.Reasoning`.
- `cmd/kubernautagent/llm_builder.go` only read `cfg.Reasoning.CapabilityOverride` (for the OpenAI-compatible replay-mode auto-detection), never `.Enabled`/`.BudgetTokens`.
- `anthropicfamily.Client` had no construction-time option to bake in a default; it only honored a per-call `Options.Reasoning`, which nothing ever set.

Net effect: `ai.llm.reasoning.enabled: true` parsed, validated, and hot-reloaded correctly, but never caused a real Anthropic/Vertex API call to include the `thinking` request parameter — a structural no-op for the exact provider (Claude) that motivated BR-AI-086 (OLS-3442).

This was fully masked by test doubles:
- `E2E-KA-AUDIT-001`'s mock-llm unconditionally returns `reasoning_content` regardless of the incoming request, so it proves capture/audit-surfacing but not that reasoning was ever *requested*.
- That E2E test only exercises the OpenAI-compatible provider path (`provider: "openai"`) — Anthropic's request-side `thinking` param was never exercised by anything above the unit-test tier (hand-authored fixtures only).

## Fix

- Add `anthropicfamily.WithReasoning(llm.ReasoningRequest)`, a construction-time option matching the documented contract on `llm.ReasoningRequest` ("resolved once at client construction, never threaded per-call from business logic"). `Client.buildParams` now falls back to this default whenever a per-call `Options.Reasoning` is nil; an explicit per-call value still wins.
- Wire `cfg.Reasoning` into both the `anthropic` and `vertex_ai` construction paths in `llm_builder.go` via a new `anthropicReasoningOptions` helper.
- Document (via code comment) why the OpenAI-compatible path intentionally does *not* thread `.Enabled`/`.BudgetTokens`: that wire protocol has no request-side reasoning toggle — capture there is correctly unconditional/model-driven (DeepSeek-reasoner-class models).

## Test plan

- [x] New UT-KA-1578-201..204 in `pkg/kubernautagent/llm/anthropicfamily/thinking_1580_test.go`: construction-time default applies when no per-call override is set, no regression when `WithReasoning` isn't used, per-call override still wins, `Enabled:false` never sends `thinking`.
- [x] New wiring tests in `cmd/kubernautagent/llm_builder_wiring_1580_1581_test.go`: `anthropicReasoningOptions` config→option mapping, and `buildAnthropicNativeClient` dispatch with a reasoning-enabled config.
- [x] `go build ./...`, `go vet ./...` clean.
- [x] `golangci-lint run` clean on changed packages.
- [x] Full `go test ./pkg/kubernautagent/... ./cmd/kubernautagent/... ./internal/kubernautagent/... ./pkg/shared/llm/... ./pkg/shared/types/...` passes, no regressions.

### DIFF
--- a/cmd/kubernautagent/ds_transport_test.go
+++ b/cmd/kubernautagent/ds_transport_test.go
@@ -18,39 +18,31 @@ package main
 
 import (
 	"net/http"
-	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 
 	"github.com/jordigilh/kubernaut/pkg/shared/types"
 )
 
-func TestBuildDSBaseTransport_WithCAFile(t *testing.T) {
-	caPath := generateTestCACert(t, "DS Test CA")
+var _ = Describe("buildDSBaseTransport", func() {
+	It("returns a custom non-default transport when a valid CA file is set", func() {
+		caPath := generateTestCACert(GinkgoTB(), "DS Test CA")
 
-	transport, err := buildDSBaseTransport(caPath, types.LLMCircuitBreaker{})
-	if err != nil {
-		t.Fatalf("expected no error with valid CA file, got: %v", err)
-	}
-	if transport == nil {
-		t.Fatal("expected non-nil transport when caFile is set")
-	}
-	if transport == http.DefaultTransport {
-		t.Fatal("expected custom transport, got http.DefaultTransport")
-	}
-}
+		transport, err := buildDSBaseTransport(caPath, types.LLMCircuitBreaker{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(transport).NotTo(BeNil())
+		Expect(transport).NotTo(BeIdenticalTo(http.DefaultTransport))
+	})
 
-func TestBuildDSBaseTransport_EmptyCAFile(t *testing.T) {
-	transport, err := buildDSBaseTransport("", types.LLMCircuitBreaker{})
-	if err != nil {
-		t.Fatalf("expected no error with empty caFile, got: %v", err)
-	}
-	if transport == nil {
-		t.Fatal("expected non-nil transport (default fallback)")
-	}
-}
+	It("returns a non-nil default-fallback transport when caFile is empty", func() {
+		transport, err := buildDSBaseTransport("", types.LLMCircuitBreaker{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(transport).NotTo(BeNil())
+	})
 
-func TestBuildDSBaseTransport_InvalidCAFile(t *testing.T) {
-	_, err := buildDSBaseTransport("/nonexistent/ca.crt", types.LLMCircuitBreaker{})
-	if err == nil {
-		t.Fatal("expected error with invalid CA file path, got nil")
-	}
-}
+	It("returns an error for an invalid CA file path", func() {
+		_, err := buildDSBaseTransport("/nonexistent/ca.crt", types.LLMCircuitBreaker{})
+		Expect(err).To(HaveOccurred())
+	})
+})
--- a/cmd/kubernautagent/fleet_readiness_wiring_test.go
+++ b/cmd/kubernautagent/fleet_readiness_wiring_test.go
@@ -19,10 +19,11 @@ package main
 import (
 	"context"
 	"net/http/httptest"
-	"testing"
 	"time"
 
 	"github.com/go-logr/logr"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 
 	kaconfig "github.com/jordigilh/kubernaut/internal/kubernautagent/config"
 	"github.com/jordigilh/kubernaut/pkg/fleet/mcpclient"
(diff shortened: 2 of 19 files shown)


jordigilh:
Added [DD-LLM-007](https://github.com/jordigilh/kubernaut/blob/fix/reasoning-request-side-wiring-1578/docs/architecture/decisions/DD-LLM-007-af-ka-anthropic-client-divergence.md), documenting why AF and KA intentionally keep separate Anthropic clients (AF is ADK-based; KA is framework-isolated per DD-HAPI-019-001) rather than unifying — this came up while scoping E2E coverage for the reasoning-request wiring fix in this PR. No AF change is needed for the #1578 bug itself (AF was never in scope for BR-AI-086).

Next: adding minimal E2E coverage in mock-llm for the actual Anthropic request-side wiring this PR fixes.

jordigilh:
Fixed the test-convention issue you flagged on `cmd/kubernautagent/llm_builder_wiring_1580_1581_test.go` (`fe224df...`): those tests used plain `testing.T`/`t.Run`, not Ginkgo/Gomega. That file predates this PR (added in #1598 by #1581), and I extended it with more `testing.T` tests instead of converting it — compounding the violation. Fixed now:

- Converted the whole file to `Describe`/`It`/`DescribeTable`.
- Added `cmd/kubernautagent/suite_test.go` — the package's first Ginkgo `RunSpecs` bootstrap (there wasn't one; the other 11 `*_test.go` files in this package still use plain `testing.T`, which is pre-existing, unrelated tech debt — happy to file a follow-up issue to convert those too if you want it tracked).
- Added a new spec (**IT-KA-1578-003**) that closes a gap the *original* test comments themselves called out: the two prior tests proved (a) `cfg.Reasoning` → `anthropicfamily.Option` mapping, and (b) `buildAnthropicNativeClient` dispatches without error — but never both together against a live HTTP call. The new spec builds a client the same way `buildAnthropicNativeClient` does internally (`anthropicReasoningOptions(cfg)` + a test-only `WithSDKOptions` base-URL override) and asserts the real outgoing JSON body contains `"thinking"`. This is a genuine live wire-level proof of the request-side fix, not just construction-time proof.

On **E2E**: this new spec is IT-tier (in-process `httptest.Server`, no Kind cluster, no separate `mock-llm` process) — it's not literally the project's E2E tier. A true E2E for this fix would need `mock-llm` to speak the Anthropic Messages API (`/v1/messages`), which it doesn't today (it only implements OpenAI-compatible endpoints — see `test/services/mock-llm/handlers/router.go`). Adding that is a real option but a separate, larger piece of work than this bugfix. Let me know if you want that scoped as a follow-up now or tracked as an issue — didn't want to bundle an unrelated mock-llm feature into this hotfix without confirming.

All of `cmd/kubernautagent` builds, vets, lints (0 issues), and passes (`go test ./cmd/kubernautagent/...`, including the pre-existing `testing.T` tests alongside the new Ginkgo suite).

jordigilh:
Closing out the E2E question raised above. Decision: **IT-tier coverage is sufficient; no mock-llm Anthropic endpoint or second KA Kind deployment needed.**

Rationale: an "E2E" test through a Kind-deployed KA + a mock-llm Anthropic endpoint would still just be testing against our own mock — the same trust boundary as the IT tests already added, just with far more infrastructure ceremony (new Deployment, new NodePort, DD-TEST-001 revision, permanent extra pod on every KA E2E run) wrapped around it. It wouldn't add real assurance about actual Anthropic API compatibility, and no other provider path here is validated against the real upstream API either (`E2E-KA-AUDIT-001` also runs against mock-llm, not real OpenAI).

The full request→response→audit loop for the #1578 Anthropic fix is already closed across existing + new tests, with no gap:
1. **Config → dispatch** (`provider: anthropic` + `reasoning.enabled: true` → `buildAnthropicNativeClient`): `IT-KA-1578-001/002`, `IT-KA-1580-001`.
2. **Dispatch → real outgoing wire request** (the actual bug this PR fixes): `IT-KA-1578-003` (new, full fidelity — real HTTP call, asserts `"thinking"` on the wire), plus `UT-KA-1580-103..106` (tier detection) and `UT-KA-1578-201..204` (`WithReasoning` default/override semantics).
3. **Incoming Anthropic response → `Message.Reasoning`**: already covered by pre-existing `UT-KA-1580-107/108/109` (visible thinking, redacted thinking, no-thinking).
4. **Multi-turn replay ordering** (Anthropic-specific: thinking block must precede tool_use): pre-existing `UT-KA-1580-110/111`.
5. **`Message.Reasoning` → audit trail → DataStorage → correlation_id reconstruction**: this half of the pipeline (`toReasoningSummary`, the gate-retry fix, audit emission) is provider-agnostic — it only operates on the shared `llm.ReasoningBlock` type, no provider branching — so `E2E-KA-AUDIT-001` (OpenAI-mocked path) already proves it end-to-end for any provider that correctly populates `Message.Reasoning`.

Also included in this PR now: converted `cmd/kubernautagent`'s remaining 11 pre-existing `testing.T`-based test files to Ginkgo/Gomega (AGENTS.md's zero-tolerance BDD mandate) — unrelated to #1578 but flagged during review, so cleaned up here rather than deferred. 61 specs pass, `go vet`/`golangci-lint` clean.

PR #1628, opened 2026-07-08
feat(helm): close apifrontend/fleetmetadatacache hardening gaps (PR A of chart refactor)

jordigilh:
## Summary

First of a three-PR sequence (A/B/C) closing chart-refactor/hardening gaps
identified while auditing the Helm chart against the Kubernaut Operator's
defaults (`kubernaut-operator/internal/resources/`). This PR (A) only closes
gaps in two services that were missing wiring every other service already
had — no new default behavior, no opt-in-by-default flips (that's PR B).

`apifrontend` and `fleetmetadatacache` were the only two mesh services
missing:
- `imagePullSecrets`
- `podSecurityContext` / `containerSecurityContext` (fleetmetadatacache had
  these hardcoded inline, not overridable via values)
- `affinity`
- scheduling (`nodeSelector` / `tolerations`) — fleetmetadatacache didn't
  even have these in the schema
- PDB support (`pdb.yaml` only had 11 of the 13 services)

## Changes

- **`apifrontend.yaml`**: wire `kubernaut.imagePullSecrets` /
  `podSecurityContext` / `affinity` / `scheduling` /
  `containerSecurityContext` into the Deployment. The container previously
  had **no** `securityContext` at all.
- **`fleetmetadatacache.yaml`**: replace the hand-rolled, non-overridable
  `securityContext` block with the same helper calls used everywhere else
  in the chart — rendered defaults are **byte-identical** (zero behavior
  change for existing users), just now overridable. Add the same
  affinity/scheduling/imagePullSecrets wiring. Append `kubernaut.labels`
  to all of its resources' `metadata.labels` (its `app.kubernetes.io/name`
  selector scheme is untouched — only non-selector labels changed).
- **`pdb.yaml`**: extend the shared PDB loop to carry a per-component
  selector-label key, so `fleetmetadatacache`'s
  `app.kubernetes.io/name` selector scheme can share the same template as
  every `app: <name>` service. Add `apifrontend` + `fleetmetadatacache`
  entries — `pdb.enabled` defaults to `false` for both, same as every
  other service today.
- **`values.schema.json` / `values.yaml`**: add `podSecurityContext`,
  `containerSecurityContext`, `pdb`, `affinity`,
  `topologySpreadConstraints`, `nodeSelector`, `tolerations` to both
  services (fleetmetadatacache was missing `nodeSelector`/`tolerations`
  entirely, not just the pod-hardening fields).
- **Bugfix (found via new test coverage)**: `apifrontend.yaml`'s
  per-persona `ClusterRole` labels had a hardcoded
  `app.kubernetes.io/part-of: kubernaut` that duplicated the same key
  already emitted by `kubernaut.labels`, producing an invalid duplicate
  YAML mapping key. This is pre-existing on `main` (unrelated to this PR's
  scope) — it was invisible because no `helm-unittest` suite previously
  covered `apifrontend.yaml`. Fixed by dropping the redundant hardcoded
  line.
- **`tests/apifrontend_fmc_hardening_test.yaml`** (new, 10 tests): default
  pod/container security context, `imagePullSecrets` rendering,
  affinity/scheduling overrides, fleetmetadatacache's rendered
  security context being unchanged, label consistency, and PDB
  selector-scheme correctness for both services (including the
  enabled-gate for fleetmetadatacache's PDB).
- **README**: documented that APIFrontend and FleetMetadataCache now
  accept the same common controller parameters as every other service.

## Test plan

- [x] `helm lint` (default `tls.mode`)
- [x] `helm template` across all three `tls.mode` values (`hook`,
  `cert-manager`, `manual`) — clean render, no errors
- [x] `python -m json.tool values.schema.json` — valid JSON
- [x] Full `helm-unittest` suite: **37/37 passing** (27 pre-existing + 10
  new)
- [x] Verified `apifrontend`'s Go binary has no production (non-test) code
  paths writing to disk, before enabling `readOnlyRootFilesystem: true`
  for the first time on that container
- [x] Verified fleetmetadatacache's rendered `securityContext` is
  byte-identical before/after (confirmed via `helm template` diff)

### DIFF
--- a/charts/kubernaut/README.md
+++ b/charts/kubernaut/README.md
@@ -622,6 +622,14 @@ If neither an override nor discovery succeeds, `helm install`/`upgrade` fails wi
 
 All controllers accept: `replicas`, `resources`, `pdb.{enabled,minAvailable,maxUnavailable}`, `podSecurityContext`, `containerSecurityContext`, `nodeSelector`, `tolerations`, `affinity`, `topologySpreadConstraints`.
 
+**APIFrontend** and **FleetMetadataCache** also accept the full set above (`podSecurityContext`,
+`containerSecurityContext`, `affinity`, `topologySpreadConstraints`, `pdb`, `nodeSelector`,
+`tolerations`) — previously these two services were missing wiring for the pod-hardening and
+scheduling helpers that every other service already used, even though the schema and values
+already implied they were supported. `pdb.enabled` defaults to `false` for both, matching every
+other service; set `apifrontend.pdb.enabled=true` / `fleetmetadatacache.pdb.enabled=true` (plus
+`minAvailable`/`maxUnavailable`) to opt in.
+
 ### Kubernaut Agent Security Hardening (BR-PLATFORM-005)
 
 KubernautAgent is the highest-risk component in the mesh — it is LLM-driven and carries the
--- a/charts/kubernaut/templates/apifrontend/apifrontend.yaml
+++ b/charts/kubernaut/templates/apifrontend/apifrontend.yaml
@@ -222,7 +222,6 @@ kind: ClusterRole
 metadata:
   name: kubernaut-tool-{{ $persona }}
   labels:
-    app.kubernetes.io/part-of: kubernaut
     kubernaut.ai/component: tool-authorization
     {{- include "kubernaut.labels" $ | nindent 4 }}
 rules:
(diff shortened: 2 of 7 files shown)


jordigilh:
Superseded — these changes were folded into #1625 at the user's request (cherry-picked as commit 8af8fc9b5), since they add up on top of the other chart changes already in that PR. Closing this standalone PR; the branch `feature/chart-gap-apifrontend-fmc` can be deleted.

PR #1629, opened 2026-07-08
feat(helm): default anti-affinity + PDB on by default (PR B, DD-PLATFORM-004)

jordigilh:
## Summary

Second of a three-PR sequence (A/B/C) closing chart-refactor/hardening gaps found while
auditing the Helm chart against the Kubernaut Operator's defaults. **Stacked on #1625**
(base branch `feature/1617-gitops-cicd-smoke-tests`, which already includes PR A's
`apifrontend`/`fleetmetadatacache` hardening) since this PR updates the `kubernaut.affinity`
helper's call sites, including the two added by PR A — rebase/merge order: #1625 first, then
this PR.

Closes the remaining Operator-parity gap: the Kubernaut Operator ships **unconditional** soft
pod anti-affinity and a PDB (`maxUnavailable: 1`) for every component
(`kubernaut-operator/internal/resources/deployments.go`, `pdb.go`); the Helm chart shipped
both as opt-in (`enabled: false`, or missing entirely pre-PR-A).

See **[DD-PLATFORM-004](docs/architecture/decisions/DD-PLATFORM-004-chart-default-hardening.md)**
for the full alternatives analysis and a known, accepted limitation.

## Changes

- **`kubernaut.affinity` helper** (`_helpers.tpl`): now injects a default soft
  (`preferred`, weight 100) pod anti-affinity term — `topologyKey:
  kubernetes.io/hostname`, `matchLabels` = the component's own pod-selector labels —
  deep-merged (Sprig `merge`) with any user-supplied `<service>.affinity` override. User
  keys win on conflict; sibling keys (e.g. `nodeAffinity`) merge in additively. Signature
  changed from `(include "kubernaut.affinity" .Values.<service>)` to `(include
  "kubernaut.affinity" (dict "component" .Values.<service> "matchLabels" (dict "app"
  "<name>")))`.
- **All 15 call sites** (13 pre-existing + `apifrontend`/`fleetmetadatacache` from PR A)
  updated to pass their actual pod-selector `matchLabels` (`app: <name>` for 14 services;
  `app.kubernetes.io/name: fleetmetadatacache` for the one service using that scheme —
  verified against each Deployment's real `spec.selector.matchLabels`, not assumed).
- **`values.yaml`**: `pdb.enabled` flipped `false` → `true` with `maxUnavailable: 1`
  (previously commented out) for the 12 services that had it disabled.
- **Bugfix**: `datastorage.pdb` was already `enabled: true` by default, but used
  `minAvailable: 1`. With `datastorage.replicas: 1`, that means **zero voluntary
  disruptions are ever permitted** (`PodDisruptionBudgetAtLimit`) — silently blocking
  `kubectl drain` / cluster-autoscaler / node upgrades indefinitely. Changed to
  `maxUnavailable: 1`, matching the Operator's own explicit test guard against this exact
  failure mode (`pdb_test.go`: `"PDB should not have MinAvailable set (causes
  PodDisruptionBudgetAtLimit with 1 replica)"`).
- **`values.schema.json`**: `#/definitions/pdb.properties.enabled.default` changed
  `false` → `true` to match.
- **New DD-PLATFORM-004** documents the decision and a known, accepted limitation found
  during testing: an explicit empty-list or `null` override of the default anti-affinity
  term is silently discarded by `mergo` (treated as an unset zero-value, so the default
  wins back) — a full override requires the user to supply their own **non-empty**
  `preferredDuringSchedulingIgnoredDuringExecution` list, which does correctly replace the
  default. Accepted rather than worked around because the injected term is `preferred`
  (soft) and therefore never blocks scheduling regardless.
- **README**: "Common Controller Parameters" section updated with the new defaults and
  override/opt-out mechanics.
- **`tests/default_hardening_test.yaml`** (new, 7 tests): default anti-affinity rendering
  for both selector schemes (`app` / `app.kubernetes.io/name`), merge-with-user-`nodeAffinity`
  behavior, full-replacement override behavior, default PDB `maxUnavailable: 1` (including a
  `datastorage` regression guard for the `minAvailable` bug above), and per-service PDB
  opt-out.
- Updated PR A's existing test (`apifrontend_fmc_hardening_test.yaml`) for the new
  PDB-enabled-by-default baseline (it previously assumed all PDBs were disabled by default).

## Test plan

- [x] `helm lint` (default `tls.mode`)
- [x] `helm template` across all three `tls.mode` values (`hook`, `cert-manager`, `manual`)
  — clean render, no errors
- [x] `python -m json.tool values.schema.json` — valid JSON
- [x] Full `helm-unittest` suite: **68/68 passing** (61 from #1625 + 7 new)
- [x] Isolated scratch-chart spike confirming the exact `merge`/`mergo` semantics for: no
  override, empty-list override (fails to opt out — documented as a known limitation),
  sibling-key override (merges in), and non-empty full-replacement override (works)
- [x] Verified each of the 15 `matchLabels` values against the corresponding Deployment's
  actual `spec.selector.matchLabels` (not assumed from naming convention)

### DIFF
--- a/charts/kubernaut/README.md
+++ b/charts/kubernaut/README.md
@@ -626,9 +626,25 @@ All controllers accept: `replicas`, `resources`, `pdb.{enabled,minAvailable,maxU
 `containerSecurityContext`, `affinity`, `topologySpreadConstraints`, `pdb`, `nodeSelector`,
 `tolerations`) — previously these two services were missing wiring for the pod-hardening and
 scheduling helpers that every other service already used, even though the schema and values
-already implied they were supported. `pdb.enabled` defaults to `false` for both, matching every
-other service; set `apifrontend.pdb.enabled=true` / `fleetmetadatacache.pdb.enabled=true` (plus
-`minAvailable`/`maxUnavailable`) to opt in.
+already implied they were supported.
+
+#### Default anti-affinity and PDB (DD-PLATFORM-004)
+
+Every service that renders a Deployment gets a default **soft** (preferred, not required)
+pod anti-affinity spreading its replicas across nodes, and `pdb.enabled` defaults to
+`true` (`maxUnavailable: 1`) for every service in `templates/pdb.yaml` — matching the
+Kubernaut Operator's defaults. Both are safe with the chart's default `replicas: 1`: a
+preferred anti-affinity term is a no-op with no peer replica to avoid, and
+`maxUnavailable: 1` (never `minAvailable`) always permits the sole pod to be voluntarily
+evicted (e.g. `kubectl drain`) rather than blocking it. Override per-service via
+`<service>.affinity` (deep-merged with the default: a sibling key like `nodeAffinity`
+merges in alongside the default `podAntiAffinity`; to fully replace the default
+anti-affinity term itself, provide your own **non-empty**
+`podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution` list — an *empty*
+list or `null` is treated as unset and the default silently wins back, see
+DD-PLATFORM-004) and `<service>.pdb.enabled=false` to opt out of the PDB entirely.
+`console`'s and `fleetmetadatacache`'s PDBs still only render when the service itself
+is enabled.
 
 ### Kubernaut Agent Security Hardening (BR-PLATFORM-005)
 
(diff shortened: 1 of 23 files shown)


jordigilh:
Superseded — these changes were folded into #1625 at the user's request (cherry-picked as commit 35076b128), since all three chart-refactor PRs (A/B/C) build on top of each other and should ship together. Closing this standalone/stacked PR; the branch `feature/chart-default-antiaffinity-pdb` can be deleted.

PR #1630, opened 2026-07-08
fix(workflowexecution): eliminate duplicate terminal audit writes on conflict retries (#1597)

### LINKED ISSUE #1597: AtomicStatusUpdate: RetryOnConflict-wraps-audit-call ordering can cause duplicate audit writes on conflict retry
## Background

Discovered during the audit-count preflight for #1572 (BR-WE-019 / DD-WE-008), while verifying that WorkflowExecution status-update paths emit exactly one audit event per logical status transition.

## Problem

`AtomicStatusUpdate` (WorkflowExecution controller, status-update helper) wraps the whole status-update-plus-audit-emission logic inside a `client-go` `retry.RetryOnConflict` closure. If:

1. The status update inside the closure succeeds against the in-memory object,
2. The audit call inside the same closure fires (recording the transition),
3. But the subsequent `Status().Update()` call to the API server fails with a conflict (e.g. another controller/reconcile loop updated the resource concurrently),

then `RetryOnConflict` will re-run the *entire* closure — including a second audit call — even though the audit event for that logical transition was already recorded once. This risks duplicate audit writes for a single real-world state transition, which is a data-quality concern for BR-AUDIT-005 (SOC2 CC8.1 audit completeness / accurate reconstruction from `correlation_id`).

This is a **pre-existing, WFE-specific architectural gap** — not introduced by #1572 — and was not fixed there since it needs its own design review (see options below) rather than a mechanical patch.

## Candidate approaches (for design review, not yet decided)

- **Option A**: Move the audit call outside the `RetryOnConflict` closure entirely, firing it once after the status update definitively succeeds (based on the final, post-retry object state) — likely the cleanest fix, but requires confirming no downstream logic depends on the audit call happening with the same closure-local variables/state as the (possibly retried) update.
- **Option B**: Make the audit call idempotent (e.g. dedupe by a stable event key such as `resource_version` + transition type) at the audit-write layer, tolerating duplicate emission attempts.
- **Option C**: Detect-and-skip re-emission inside the closure on retry (e.g. only audit on the first attempt), which requires tracking attempt state across retries.

## Scope

- WorkflowExecution controller's `AtomicStatusUpdate` (or equivalently named status-update-with-audit helper)
- Any other controllers using the same wrap-audit-in-RetryOnConflict pattern should be audited for the same gap once the fix approach is decided

## References

- DD-WE-008 (where this was surfaced, in the context of BR-WE-019 audit-count preflight)
- BR-AUDIT-005 v2.0 (audit completeness / reconstruction requirement)
- Issue #1572 (where this was discovered but explicitly not bundled)

## Acceptance Criteria

- [ ] Design decision made and documented (DD-XXX or equivalent) on which approach to take
- [ ] Fix implemented with a regression test proving no duplicate audit event is emitted when a status update is retried after a conflict
- [ ] Other controllers using the same pattern are checked for the same gap

jordigilh:
## Summary

- `MarkCompleted`/`MarkFailed`/`MarkFailedWithReason` recorded the terminal audit event (`workflow.completed`/`workflow.failed`) **inside** the `AtomicStatusUpdate` closure. Since `RetryOnConflict` re-runs the entire closure on every optimistic-lock conflict, a single logical transition that hit even one conflict produced duplicate audit events — violating BR-AUDIT-005/SOC2 CC8.1 trace integrity.
- Extracted the audit write + `AuditRecorded` condition into a new `recordTerminalAudit` helper, called **after** the primary `AtomicStatusUpdate` durably commits (not from inside its retryable closure), aligning WorkflowExecution with the existing AIAnalysis/SignalProcessing precedent.
- Adds `DD-WE-009` documenting the decision, rejected alternatives, and spike validation.

Closes #1597.

## Test plan

- [x] `UT-WE-1597-001/002/003`: new conflict-injection regression tests (via `interceptor.Funcs`) proving exactly one audit event is recorded per logical transition despite a forced `Status().Update()` conflict, for `MarkCompleted`, `MarkFailed`, and `MarkFailedWithReason` respectively.
- [x] Verified RED: all 3 tests fail against the pre-fix code (audit count scales with conflict retries).
- [x] Verified GREEN: all 3 tests pass against the fix.
- [x] Full unit suite (`pkg/workflowexecution/...`, `internal/controller/workflowexecution/...`) — pass, zero regressions.
- [x] Full integration suite (`make test-integration-workflowexecution`) — 108/108 pass, including `AuditRecorded`-condition assertions in `conditions_integration_test.go`/`job_lifecycle_integration_test.go`.
- [x] `go build ./...`, `go vet ./...`, `golangci-lint run` — clean.

### DIFF
--- a/docs/architecture/decisions/DD-WE-009-terminal-audit-ordering.md
+++ b/docs/architecture/decisions/DD-WE-009-terminal-audit-ordering.md
@@ -1,0 +1,247 @@
+# DD-WE-009: Terminal Audit Write Ordering (Duplicate Audit Fix)
+
+**Version**: 1.0
+**Date**: 2026-07-07
+**Status**: APPROVED
+**Author**: WorkflowExecution Team
+**Reviewers**: Platform Team
+
+---
+
+## Context
+
+`MarkCompleted`, `MarkFailed`, and `markFailedInternal` (the shared core of
+`MarkFailedWithReason`/`MarkFailedAsDeduplicated`) each perform a terminal phase
+transition via `StatusManager.AtomicStatusUpdate`, which wraps the caller's closure in
+`k8sretry.RetryOnConflict`:
+
+```go
+func (m *Manager) AtomicStatusUpdate(ctx context.Context, wfe *workflowexecutionv1alpha1.WorkflowExecution, updateFunc func() error) error {
+    return k8sretry.RetryOnConflict(k8sretry.DefaultRetry, func() error {
+        if err := m.client.Get(ctx, client.ObjectKeyFromObject(wfe), wfe); err != nil { // refetch
+            return fmt.Errorf("failed to refetch WorkflowExecution: %w", err)
+        }
+        if err := updateFunc(); err != nil { // caller's closure
+            return fmt.Errorf("failed to apply status updates: %w", err)
+        }
+        if err := m.client.Status().Update(ctx, wfe); err != nil { // may return a conflict
+            return fmt.Errorf("failed to atomically update status: %w", err)
+        }
+        return nil
+    })
+}
+```
+
+Prior to this decision, the three transition helpers (`applyCompletedTransition`,
+`applyFailedStatusTransition`, and `markFailedInternal`'s inline closure) called the
+`AuditManager.RecordWorkflow{Completed,Failed}` audit write and set the `AuditRecorded`
+condition **inside** this closure. Because `RetryOnConflict` re-runs the **entire**
+closure — refetch, phase transition, and audit call — on every optimistic-lock conflict,
+a single logical terminal transition that hit even one conflict produced **N audit
+events for N closure executions**, not one. This is [issue #1597](https://github.com/jordigilh/kubernaut/issues/1597),
+originally surfaced as a follow-up item from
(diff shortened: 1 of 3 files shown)


jordigilh:
## CI status update

Build, Lint, and all Unit Tests relevant to this PR's change passed cleanly (`Unit Tests (workflowexecution)`, `Build (workflowexecution/*)`). Integration and E2E jobs were skipped due to fail-fast gating on unit-test success across the monorepo.

The only failure blocking `Merge Gate` is **`Unit Tests (kubernautagent)`**, on `UT-KA-1390-027`, which is an intermittent, pre-existing race condition in `internal/kubernautagent/session` — a package this PR does not touch (verified via `git diff`). Root-caused and filed as #1631; not fixed here since it is out of scope for #1597 and touches a security-sensitive code path (interaction with the `IR-4.a` security-escalation guard) that deserves its own review.

Leaving this PR as-is (unmerged, red) pending a decision on #1631. Re-running the failed job may pass on a favorable goroutine-scheduling ordering, but per project policy flaky/racy failures should be root-caused and fixed rather than retried.

PR #1709, opened 2026-07-22
fix(security): default postgres/valkey to restricted PSA contexts

### LINKED ISSUE #1685: Helm chart: postgresql/valkey pods don't default to restricted-PSA-compliant securityContext
## Context

During an EKS/AKS deployment-gap evaluation of the Helm chart (`charts/kubernaut/`), we found that the bundled `postgresql` and `valkey` Deployments do not default to Pod Security Admission (PSA) "restricted"-profile-compliant `securityContext` settings, unlike every Kubernaut-authored service in the chart.

**Scope note**: `postgresql`/`valkey` in this chart are dev/eval-convenience components only — production deployments are already documented as BYO HA Postgres/Valkey (`charts/kubernaut/README.md`, "Production Configuration" section), deployed and hardened independently of this chart. So this is **not a production blocker**, but it does affect any quick-start/eval install into a namespace that enforces `pod-security.kubernetes.io/enforce: restricted` — a common default posture on managed clusters (EKS, AKS).

## Finding

All Kubernaut-authored services (gateway, apifrontend, aianalysis, etc.) render their `securityContext` via the `kubernaut.podSecurityContext` / `kubernaut.containerSecurityContext` helpers, which default to `runAsNonRoot: true`, `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`, `capabilities.drop: [ALL]` — restricted-PSA-compliant out of the box.

`postgresql` and `valkey` instead use a separate, looser helper:

```705:712:charts/kubernaut/templates/_helpers.tpl
Generic merge(override, defaults) -> toYaml for a securityContext block,
for components needing different defaults than kubernaut.podSecurityContext/
kubernaut.containerSecurityContext above (e.g. postgresql/valkey, whose
upstream images already run non-root and need write access to their own
data directories, so they don't set runAsNonRoot/readOnlyRootFilesystem/
capabilities.drop by default).
```

```61:70:charts/kubernaut/templates/infrastructure/postgresql.yaml
      securityContext:
        {{- include "kubernaut.mergedSecurityContext" (dict "override" .Values.postgresql.podSecurityContext "defaults" (dict "seccompProfile" (dict "type" "RuntimeDefault"))) | nindent 8 }}
      ...
          securityContext:
            {{- include "kubernaut.mergedSecurityContext" (dict "override" .Values.postgresql.containerSecurityContext "defaults" (dict "allowPrivilegeEscalation" false)) | nindent 12 }}
```

(identical pattern in `valkey.yaml:61-70`.) Neither sets `runAsNonRoot: true` nor `capabilities.drop: ["ALL"]` by default. Under a namespace enforcing `pod-security.kubernetes.io/enforce: restricted`, these two Deployments would be rejected at admission unless the operator explicitly supplies `postgresql.podSecurityContext` / `postgresql.containerSecurityContext` (and the `valkey.*` equivalents).

The override path already exists and is schema-validated:

```423:424:charts/kubernaut/values.schema.json
        "podSecurityContext": { "$ref": "#/definitions/securityContext" },
        "containerSecurityContext": { "$ref": "#/definitions/securityContext" },
```

— it's just not defaulted-on, and not called out anywhere in `values.yaml` comments or `README.md`.

## Proposed fix

One of:
1. Default `postgresql`/`valkey` to restricted-compliant settings (`runAsNonRoot: true`, `capabilities.drop: [ALL]`) where the upstream images tolerate it, keeping `readOnlyRootFilesystem: false` if needed for their data directories (verify per-image).
2. If (1) isn't feasible for one or both images, at minimum: document the `podSecurityContext`/`containerSecurityContext` override in `values.yaml` comments and `README.md`, with a copy-pasteable restricted-PSA example.

## Related, not in scope here

This evaluation also raised the question of whether to keep bundling `postgresql`/`valkey` in the chart at all going forward, given production users are already expected to BYO HA infra for both. That's a separate design conversation (potential chart-scope reduction) and shouldn't block this ticket — if it happens, this hardening work becomes moot, but until then the dev-convenience path should not silently fail on a restricted-PSA namespace.

## Priority

P2 — affects dev/eval installs only, not production (which already bypasses these components via BYO infra guidance).

kevinnft:
## Summary
Closes #1685.

Bundled `postgresql` and `valkey` Deployments previously only defaulted to `seccompProfile: RuntimeDefault` / `allowPrivilegeEscalation: false`. Under a namespace with `pod-security.kubernetes.io/enforce: restricted`, they would be rejected.

## Changes
- **Pod defaults**: `runAsNonRoot: true` + `seccompProfile.type: RuntimeDefault`
- **Container defaults**: `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `readOnlyRootFilesystem: false` (data dirs must stay writable)
- Document override knobs in `values.yaml` (`podSecurityContext` / `containerSecurityContext`)
- Helm unittest suite `infra_psa_hardening_test.yaml`

## Notes
- Production BYO Postgres/Valkey path is unchanged (`enabled: false`)
- Overrides still merge over defaults via `kubernaut.mergedSecurityContext`

## Test plan
- [x] Template defaults updated
- [x] helm-unittest file added
- [ ] `helm unittest charts/kubernaut` (if plugin available in CI)


jordigilh (changes requested):
Thanks for tackling this — the diagnosis is right (postgres/valkey are the only components bypassing the `kubernaut.podSecurityContext`/`containerSecurityContext` pattern), and I like that you kept `readOnlyRootFilesystem: false` and reused the existing `mergedSecurityContext` override mechanism instead of inventing a new one. The `helm-unittest` coverage is a good instinct too.

That said, I checked this out locally and validated it against the real chart tooling and the actual upstream images, and found a few things that need fixing before this can merge. Posting the evidence so it's easy to reproduce/confirm.

## 1. `values.schema.json` wasn't updated — the chart no longer builds

`postgresql`/`valkey` in `values.schema.json` have `"additionalProperties": false` and don't list `podSecurityContext`/`containerSecurityContext` (unlike every other service's schema block, which references `#/definitions/securityContext`). Adding those keys to `values.yaml` without updating the schema breaks the chart outright, for every install — not just restricted-PSA ones:

```
$ helm lint charts/kubernaut/
[ERROR] values.yaml: - at '/valkey': additional properties 'containerSecurityContext', 'podSecurityContext' not allowed
[ERROR] templates/: - at '/postgresql': additional properties 'containerSecurityContext', 'podSecurityContext' not allowed
Error: 1 chart(s) linted, 1 chart(s) failed

$ helm template charts/kubernaut/
Error: values don't meet the specifications of the schema(s) in the following chart(s): ...

$ helm unittest charts/kubernaut/ -f 'tests/infra_psa_hardening_test.yaml'
Tests: 5 failed, 5 errored, 0 passed, 5 total
```

`main` lints clean (0 failures), so this is a regression from this PR. It's also exactly what the `values.schema.json validation` gate (`helm lint --strict`) in `ci-pipeline.yml`'s `helm-smoke-test` job exists to catch — it just hasn't run yet on this PR.

**Fix**: add to both the `postgresql` and `valkey` definitions in `values.schema.json`:
```json
"podSecurityContext": { "$ref": "#/definitions/securityContext" },
"containerSecurityContext": { "$ref": "#/definitions/securityContext" },
```

## 2. `capabilities.drop: [ALL]` crashes both containers on every startup (reproduced)

Both `postgres:16-alpine` and `valkey/valkey:8-alpine` start as root and rely on `gosu`/`setpriv` (which need `CAP_CHOWN`/`CAP_SETUID`/`CAP_SETGID`) to fix data-dir ownership and drop to their service user. Dropping all capabilities removes exactly what those entrypoints need:

```
$ podman run -d --cap-drop=ALL -e POSTGRES_PASSWORD=test postgres:16-alpine
Exited (1)
  chmod: /var/lib/postgresql/data: Operation not permitted
  chmod: /var/run/postgresql: Operation not permitted
  error: failed switching to 'postgres': operation not permitted

$ podman run -d --cap-drop=ALL valkey/valkey:8-alpine
Exited (127)
  setpriv: setresuid failed: Operation not permitted
```

This reproduces with plain `--cap-drop=ALL` (the same primitive `capabilities.drop: [ALL]` maps to in a container's securityContext), so it isn't PSA- or Kubernetes-specific — both containers will crash-loop with these defaults regardless of namespace.

## 3. `runAsNonRoot: true` without `runAsUser` (documented risk, not independently reproduced here)

Neither image sets `USER` in its Dockerfile (confirmed via `skopeo inspect --config`: both report `User: null`), so the image's default resolved UID is root. Setting `runAsNonRoot: true` with no `runAsUser` is the well-documented Kubernetes failure mode where the kubelet rejects the container before it ever starts: `CreateContainerConfigError: container has runAsNonRoot and image will run as root` (see e.g. https://github.com/CrunchyData/postgres-operator/issues/3063 hitting the same thing with the official Postgres image). This would need a `runAsUser` default matching each image's real UID (postgres user is UID 70 in `postgres:16-alpine`; valkey user is UID 999 in `valkey/valkey:8-alpine`).

Once you add `runAsUser`, you'll likely also need to double check volume ownership (`fsGroup` and/or an init container) — bypassing the root-then-`gosu` startup means the entrypoint's own data-dir `chown` step never runs, so a freshly-provisioned PVC (commonly root-owned) may not be writable by the non-root UID depending on the storage class/CSI driver. I wasn't able to fully validate this last part on a real cluster (would want `fsGroup` + a real PVC-backed CSI volume, not just a local sandbox), so please confirm this with a real `kind`/cluster smoke test — the repo already has `scripts/helm-smoke-test.sh` for this.

## Suggested path forward
1. Update `values.schema.json` for both components.
2. Add `runAsUser` defaults (70 for postgres, 999 for valkey) alongside `runAsNonRoot: true`.
3. Verify volume writability under the new non-root UID (fsGroup or init container if needed).
4. Re-run `helm lint --strict`, `helm template`, and `helm unittest` locally — all three currently fail — and ideally a real `kind` smoke test against a namespace with `pod-security.kubernetes.io/enforce: restricted`, per the original issue.

Happy to help iterate on this — the core idea is right, it just needs the schema + runtime pieces to catch up with the intent.

kevinnft:
Thanks for the detailed review and the podman reproductions — addressed in the latest push:

1. **values.schema.json** — added `podSecurityContext` / `containerSecurityContext` (`$ref` securityContext) on both `postgresql` and `valkey`.
2. **No more `capabilities.drop: [ALL]`** on these images — defaults no longer drop caps (gosu/setpriv need CHOWN/SETUID/SETGID).
3. **`runAsUser` + `fsGroup`** defaults: postgres **70**, valkey **999**, with `runAsNonRoot: true` at pod and container level.
4. **helm-unittest** expectations updated to match (assert non-root UIDs, assert capabilities is null).

I don't have helm/kind in this environment to re-run `helm lint --strict` / `helm unittest` / smoke locally; please re-check when convenient. Happy to iterate if volume ownership still needs an init container on your CSI class.

jordigilh:
Thanks so much for your contribution!

Issue #1740, opened 2026-07-26
PROPOSAL: Replace Custom Investigation Loop with NousResearch Hermes Agent Harness


rrbanda:
# PROPOSAL: Replace Custom Investigation Loop with NousResearch Hermes Agent Harness

**Author:** Raghuram Banda  
**Date:** 2026-07-26  
**Status:** Draft (Pending Maintainer Review)  
**Target:** Kubernaut Agent (`cmd/kubernautagent`, `internal/kubernautagent/investigator/`)  
**Related:** DD-HAPI-019-001 (Framework Isolation), PROPOSAL-EXT-005/006 (superseded agent runtimes)

---

## 1. Problem Statement

The Kubernaut Agent maintains a bespoke agentic investigation loop in Go (`internal/kubernautagent/investigator/`, ~2400 lines production code across 14 files, ~30 test files). This custom framework implements:

- ReAct-pattern conversation loop (`investigator_loop.go`)
- Multi-phase orchestration: RCA → Workflow Discovery (`investigator_phases.go`, `investigator_rca.go`, `investigator_workflow_selection.go`)
- Tool dispatch with sanitization and summarization (`investigator_tools.go`)
- Validation gates with retry/self-correction (`investigator_gates.go`)
- Anomaly detection (`anomaly.go`)
- Token accumulation and context management (`token_accumulator.go`)
- Per-phase LLM client resolution (`phase_resolver.go`)

**The maintenance burden of this custom framework is significant:**

1. **LLM response parsing fragility** — The `"no JSON found in response"` failures observed in production (see agent logs 2026-07-23) are a direct consequence of hand-rolled response parsing. Each new LLM provider or model revision risks introducing format incompatibilities that require custom mitigation code.

2. **Sequential tool execution** — The current loop executes tools one-at-a-time. Parallel tool execution (which LLMs increasingly request via parallel function calls) is not supported, limiting investigation speed.

3. **No subagent delegation** — Complex incidents requiring multi-hypothesis exploration cannot spawn specialist sub-investigations. The current architecture is single-threaded-single-agent.

4. **Duplicated concerns** — Context window management, iteration budgeting, error classification, model failover, streaming, and interrupt handling are all implemented bespoke when open-source agent runtimes have solved these problems with significantly more testing surface area.

[...]

Issue #1681, opened 2026-07-19
AuthBridge: extend intercepted-traffic relay to feed shadow-evaluator (in addition to audit)


jordigilh:
## Summary

Extend `AuthBridge`'s existing traffic-interception responsibility (currently scoped to credential injection + audit relay, #1535) with a second consumer: tee a copy of the intercepted agent LLM/MCP traffic to a shadow-style security evaluator, in addition to the audit relay to KA.

This closes a known grounding gap in the current shadow-agent design: today's `LLMProxy` only submits the primary LLM's final response text (`resp.Message.Content`) to the shadow evaluator, not the underlying tool-call evidence — so the shadow agent cannot validate whether an RCA is actually grounded in the investigation's real findings vs. fabricated/injected content (it has no ground truth). Since AuthBridge already sees the complete LLM+MCP tool-call stream by construction (that's its existing credential-interception job), it is well-positioned to give the shadow evaluator full-fidelity visibility, instead of the current filtered/summarized relay.

## Scope (v1.7)

- Extend AuthBridge to tee intercepted traffic to a second destination (shadow evaluator) alongside the existing audit-relay-to-KA path (#1535)
- Evaluator identity is explicitly **not** decided by this issue: could remain Kubernaut's own bespoke shadow-agent LLM (current, per `ADR-KA-001`), or a future NemoClaw-based evaluator (see `PROPOSAL-EXT-003` DG-22 / `ADR-KA-002`) — this issue only scopes the relay/tee mechanism, not which evaluator consumes it
- AuthBridge's role as the interception point is durable regardless of that evaluator choice, and regardless of the OpenShell sandboxing migration (OpenShell = process isolation, AuthBridge = traffic interception; they coexist, not substitute)

## Explicitly out of scope

- Choosing/replacing the evaluator behind the relay (separate decision)
- OpenShell/AuthBridge sidecar coexistence validation (already tracked as untested in `SPIKE-OPENSHELL-OAS-RUNTIME.md` §8)

## References

- #1535 (AuthBridge audit-relay extension — this issue extends the same mechanism with a second consumer)
- #1536 (opaque-OCI-agent runtime-agnostic direction)
- `docs/architecture/decisions/ADR-KA-001-shadow-agent-alignment-check.md` (current shadow-agent design, in-process proxy)
- `docs/architecture/decisions/ADR-KA-002-agent-security-defense-in-depth.md` (new — defense-in-depth model this issue's Layer 1/3/4 split refers to)

Issue #1536, opened 2026-07-03
Define AgenticWorkflow CRD spec v2: OCI image ref + input/output contract (runtime-agnostic)


jordigilh:
## Summary

Define the `AgenticWorkflow` CRD spec for the runtime-agnostic direction: agents packaged as fully opaque OCI images (no Kubernaut-known runtime type, no internal spec-file format), executed via the image's own `ENTRYPOINT`. This supersedes the runtime-aware CRD sketch in `SPIKE-OCI-RUNTIME-CONTRACT.md`.

**Related**: #818 (Vision epic — externalized/pluggable agent runtime direction), #1535 (audit relay — AuthBridge extension), #883 (closed — recipe/manifest format convergence, superseded by OCI opacity)

## Why the existing sketch doesn't work anymore

The CRD sketch in `docs/architecture/spikes/oci-runtime-contract/README.md` assumed Kubernaut is runtime-aware:

```yaml
execution:
  runtime: oas  # enum: goose, oas, deepagent
  bundle: registry.example.com/kubernaut/oomkill-investigator:v1.0.0
  bundleDigest: sha256:abc123...
  runtimeConfig:
    maxToolCalls: 30
    maxToolCallsPerTool: 10
```

This modeled an ACP server reading `ai.kubernaut.entrypoint`/`ai.kubernaut.spec-version` OCI labels, extracting a spec file from inside the image, and loading it into a runtime-specific adapter (Goose CLI, OAS `AgentSpecLoader`, LangGraph `create_react_agent`). None of that applies once agents are opaque OCI images — Kubernaut never selects a runtime or looks inside the image; the image's own `ENTRYPOINT` is the entire launch contract.

That spike (and its embedded CRD sketch) has been annotated SUPERSEDED, not deleted — see `docs/architecture/spikes/SPIKE-OCI-RUNTIME-CONTRACT.md` and `docs/architecture/spikes/oci-runtime-contract/README.md` for the historical context and what generic concerns survive.

## What survives from the old sketch (reframed as generic, not runtime-specific)

- `bundle` + `bundleDigest` — standard OCI image reference + digest pin (same pattern as `RemediationWorkflow`, no custom contract needed)
- `labels` (severity/environment/component) — workflow-catalog selection metadata, orthogonal to what's inside the image
[...]

Issue #1535, opened 2026-07-03
Agent-pod audit relay: stream OCI agent LLM/tool-call traffic to KA (AuthBridge extension)


jordigilh:
## Summary

Decision: for the future runtime-agnostic agent architecture (agents packaged as OCI images, executed in an OpenShell sandbox with an `AuthBridge` sidecar for credential interception), the audit trail for the agent's LLM/tool-call communication stream will be captured by **relaying the intercepted traffic from AuthBridge to KA**, which writes to Data Storage using its existing, already-approved async-buffered audit pipeline. This was chosen over the alternative of giving a thin Data Storage client sidecar its own direct DS-write credentials.

**Related**: #818 (Vision: Kubernaut as an Intelligent Operations Platform — externalized/pluggable agent runtime direction)

Not related to #883 (that issue tracks recipe/prompt-bundle *template format* convergence — a different concern from this audit/security architecture work).

## Why "stream to KA" over "thin DS-client sidecar"

- `DD-AUDIT-001` (Audit Responsibility Pattern, 98% confidence, approved) already rejected the structurally identical pattern for Tekton `TaskRun` audit sidecars: an ephemeral, per-task component cannot guarantee audit persistence before teardown and cannot verify completeness. The same reasoning applies to an ephemeral, per-investigation agent pod.
- `DD-AUDIT-003` v1.3 already designates KA as **the sole authoritative emitter** for the `aiagent` event category. A sidecar with its own DS-write credential would either fork that authority or impersonate KA's identity — both worse than relaying through KA.
- Concentrating Data Storage write credentials in KA alone (rather than distributing them into every agent-adjacent sidecar) minimizes the population of principals that can write to the audit-of-record table — directly serving **FedRAMP AU-9** ("protection of audit information") and **AC-6** (least privilege), especially given the agent pod is the most likely component to be adversarially influenced (prompt injection, tool-response manipulation).

## Design (validated by spike — see comment below for full spike log)

- **No new sidecar container.** `AuthBridge` already terminates/intercepts the agent's outbound LLM and MCP tool traffic to inject scoped credentials — it already has full visibility into the payloads. Extend it with a second responsibility (relay), rather than standing up a second interception point.
[...]

PR #1742, opened 2026-07-26
PROPOSAL: Pluggable investigation harness (spike artifacts + proposal)

rrbanda:
## Summary

Adds proposal and validated spike artifacts for making the investigation agent harness **pluggable** -- same pattern as workflow execution engines (Tekton/Job/Ansible) but for the investigation loop.

**Principle:** Kubernaut defines the contract (inputs, tools, output schema, audit hooks). Customers choose the runtime. The current Go agent remains the default.

## What's Included

- `docs/architecture/proposals/PROPOSAL-HERMES-AGENT-HARNESS.md` -- full technical proposal
- `docs/architecture/proposals/hermes-agent-harness/` -- validated spike artifacts:
  - `SOUL.md` -- agent persona (investigation protocol, two-phase flow)
  - `config.yaml` -- runtime configuration
  - `skills/` -- K8s investigation, workflow catalog, structured output

## Spike Results (validated on OpenShift cluster)

| Metric | Result |
|--------|--------|
| Root cause identified | ImagePullBackOff (correct) |
| Previous revision found | Yes, revision 4 (no hallucination) |
| Structured output | Complete JSON matching InvestigationResult schema |
| Time | 74 seconds |

## Key Point

This is **additive** -- zero changes to existing Go agent code. The pluggable harness is an optional alternative for teams that prefer skills-as-config, Python tooling, or different model ecosystems.

## Test Plan

Spike validated on `kubernaut-hermes-spike` namespace against the same `demo-storefront/payment-api` ImagePullBackOff scenario that the existing Go agent was tested against.

Refs: #1740


jordigilh:
Thanks for pushing this into a spike with real cluster results, and for reframing it as pluggable/additive rather than a wholesale replacement.

**The underlying idea — a pluggable investigation harness — is sound, but I'd shape it differently than this PR does.** It makes sense as an isolated runtime, split out of KA entirely: the LLM/agent runs as an OCI image in a sandboxed environment (OpenShell), with AuthBridge as the auth/interception layer mediating every outbound LLM and tool call. As long as the agent fulfills KA's input/output contract and the full LLM/tool-call transcript is captured for audit, this is doable — and that's not hand-wavy: #1535 already has a spike-validated design for exactly this (AuthBridge relays intercepted traffic to KA, which writes it through the existing, already-approved audit pipeline — no new Data Storage-facing code needed). It's not shipped yet, though — #1535 is still open, with at least one explicit compliance gap flagged before GA (actor attribution: KA currently hardcodes `ActorType=Service, ActorID=kubernaut-agent` on every audit event, which would misattribute records once the call actually executes inside a third-party container).

That's not a new idea for this repo either — it's the *current, already-adopted* direction: #818 (vision epic) describes KA evolving into exactly this kind of supervised harness; #1536 (open, in progress) is defining the OCI image + input/output contract for it; ADR-KA-002 (three weeks old) formalizes the AuthBridge/OpenShell security model for it.

What I'd flag about this PR's specific artifacts: they're shaped like an earlier iteration of this idea, not the current one. `SOUL.md`/`SKILL.md`/`config.yaml` are committed directly into the docs tree as prompt/skill definitions that KA would presumably need to load and interpret — that's the `PromptBundle` pattern from `PROPOSAL-EXT-002` (April 2026), which went through two more iterations (Goose recipes, then OAS Runtime) before being explicitly superseded on July 3: *"KA no longer compiles/bridges customer prompts into any runtime-specific recipe format... Agent runtimes are now packaged as generic OCI images, opaque to KA."* The spike itself reflects this too — it's a raw `hermes -z "prompt"` CLI invocation with a `cluster-reader` ClusterRole standing in for real RBAC, not an OCI image behind AuthBridge/OpenShell.

So I'd suggest reshaping this around #1536's contract rather than the prompt-bundle shape: package the agent (Hermes or otherwise) as an opaque OCI image with its own entrypoint, run it through OpenShell, and let AuthBridge mediate its traffic and relay the transcript per #1535's design. That also resolves a couple of things I noticed in the spike as currently structured — it bypasses the existing tool-output sanitization pipeline (prompt-injection stripping, credential scrubbing) that KA's own tools already go through, and the single validated scenario's instructions read as fairly specific to that one test case rather than general investigation guidance, so I wouldn't read the "no hallucination" result as a framework-level finding yet.

Happy to help scope a spike against #1536/#1535's actual contract if there's interest — that seems like the higher-value validation than a specific-runtime pitch at this stage.

rrbanda:
## OpenShell (NVIDIA) as the AuthBridge/OpenShell Runtime

Re: your comment about reshaping this around #1536/#1535's contract (AuthBridge + sandboxed execution) —

[NVIDIA OpenShell](https://github.com/NVIDIA/OpenShell) (Apache 2.0, 7.8K stars) already implements exactly that pattern. It's purpose-built as "the safe, private runtime for autonomous AI agents" and provides:

| Kubernaut needs (#1535 / #1536) | NVIDIA OpenShell provides today |
|---|---|
| **AuthBridge** — intercept outbound LLM calls, inject credentials | **Providers** — "Credentials never leak into sandbox; injected as env vars at runtime" + **Inference routing** — "strips caller credentials, injects backend credentials, forwards to managed model" |
| **Sandboxed execution** — isolated runtime for untrusted agents | **Sandbox** — isolated container with policy-enforced egress, filesystem, and process constraints |
| **Audit capture** — full LLM/tool transcript | **Policy Engine** — logs all allowed/denied requests; gateway is the auth boundary |
| **Opaque OCI image** — agent runs its own entrypoint | **BYOC** — `openshell sandbox create --from registry.io/img:v1` (any container image) |
| **Declarative security policies** | **YAML policies** — 4 layers: filesystem, network, process, inference (hot-reloadable) |
| **Kubernetes deployment** | **Helm chart** — `helm install openshell oci://ghcr.io/nvidia/openshell/helm-chart` + OpenShift support documented |

Notably, **Hermes Agent is explicitly listed as a supported agent** in OpenShell's README alongside Claude Code, Codex, OpenCode, and Copilot — running via [NemoClaw](https://github.com/NVIDIA/NemoClaw) for managed inference.

### What this means for #1536/#1535

Rather than building AuthBridge and the sandboxed execution layer from scratch, Kubernaut could use OpenShell as the runtime substrate for the v1.7 AgenticWorkflow feature:

```
KA receives investigation request
  → Creates OpenShell sandbox from customer's agent OCI image
  → OpenShell Policy Engine enforces: network egress, filesystem access, process constraints
  → OpenShell Inference routing: strips/injects LLM credentials
  → OpenShell Gateway logs full transcript
  → KA reads result from sandbox output
  → KA writes transcript to DataStorage audit trail (existing pipeline)
```

This resolves:
- Actor attribution (#1535 compliance gap) — OpenShell gateway is the auth boundary; KA attributes audit events to the sandbox identity
- Credential isolation — OpenShell providers inject credentials without sandbox filesystem exposure
- Tool-output sanitization — OpenShell policy engine can deny outbound requests matching exfiltration patterns

### Questions

1. Does this align with the architectural direction of #1536/#1535, or are there requirements OpenShell doesn't cover?
2. Would a spike validating the Hermes investigation agent running inside OpenShell on this cluster (with policy enforcement) be a useful next step?
3. Is there concern about taking a dependency on NVIDIA's project vs building in-house?

jordigilh:
This isn't a new direction for the project — it's already the presumptive plan, and further along than a design doc: see `SPIKE-OPENSHELL-OAS-RUNTIME.md` and `ADR-KA-002`.

So, to your three questions:

**1. Does this align with #1536/#1535?** Yes — already does. The one gap that's explicitly still open in `ADR-KA-002` itself is AuthBridge/OpenShell sidecar coexistence in one pod (§8 of the spike) — flagged as "open, untested," not yet validated.

**2. Would a spike validating Hermes inside OpenShell be useful?** Yes, targeting that untested §8 gap specifically rather than re-validating what's already covered. A few things the scope needs to get right, based on where this discussion went:

- **Audit path**: doesn't need KA to land anything in Data Storage — that's #1535's separate, already-designed job. It's enough for AuthBridge to log what it intercepts and show that's a complete, accurate transcript of the agent's activity. That's the prerequisite #1535 needs anyway.
- **Don't test against the adversarial scenario set as originally suggested** — that would actually exercise NemoClaw's inference/policy layer, not the §8 coexistence gap, and nobody's evaluated NemoClaw itself yet. That surfaces two things worth naming as genuinely open rather than assuming a setup exists: **NemoClaw vs. Kubernaut's own shadow agent isn't a decision that's been made** (today's shadow agent per `ADR-KA-001` is KA's own bespoke LLM, in-process; NemoClaw is a named candidate per `ADR-KA-002` DG-22, not a choice); and **#1681** (open, v1.7) already proposes teeing AuthBridge's intercepted traffic to a shadow evaluator alongside the audit relay — for the same reason: today's shadow agent only sees final response text, not tool-call evidence, so it can't ground-check an RCA. Worth noting: the audit-log capture above would double as groundwork for both #1535 and #1681.
- **Secret/PII redaction on tool output has no answer today, and this one's concrete, not speculative.** KA's own loop already does this — every tool result passes through a sanitization pipeline (`internal/kubernautagent/investigator/investigator_tools.go`) that redacts credentials, K8s `Secret` data, and PII before the LLM sees it, failing closed on error for SOC2. For an opaque OCI agent, KA can't reach into the agent's own tool-calling code, so this has to happen at AuthBridge instead — and I checked: nothing in AuthBridge's shipped plugin set does this today (confirmed against the actual plugin directory, not just docs). NemoClaw doesn't fill this gap either — its scanner detects *injection attack patterns*, not legitimate secret/PII content; a plain `kubectl get secret -o json` isn't an attack, so it wouldn't be touched.
  The closest existing answer is **CPEX**, an AuthBridge build variant (`authbridge-cpex`) with real, working hooks in the right place (`cmf.tool_post_invoke` for MCP responses, `cmf.llm_output` for LLM responses) and an actual `pii-scanner` plugin, demoed end-to-end. But it's young — pre-1.0, ~7 months old, 11 stars, a 2-3 person team, and its own maintainers' recent OpenShell-integration proposal explicitly excludes PII/response-mutation from v1 scope and calls out a Rust MSRV blocker against OpenShell. A native plugin against AuthBridge's base framework is also technically viable (same language as KA, direct reuse of KA's already-proven regex rules) — but with CPEX already covering this territory, it's unlikely to be accepted upstream, meaning Kubernaut would own it out-of-tree indefinitely against a plugin framework that's itself only ~2.5 months old.
  **Suggested sequencing**: spike CPEX's `pii-scanner` first; only build a native fallback if that spike shows CPEX isn't viable.
- Separately (also confirmed while digging into this): the "just OpenShell + AuthBridge" deployment model may not hold as-is — every documented AuthBridge deployment path, including the one billed as "manual," still depends on the kagenti-operator for sidecar injection and Keycloak client registration. Not the focus here, but worth a line in the spike scope so it's not assumed away.

**3. Concern about an NVIDIA dependency vs. building in-house?** Already assessed — the spike's explicit recommendation was to proceed. Worth naming two things plainly: NVIDIA is a very different vendor-risk profile than `codeany-ai` (the two-month-old, unknown-provenance org whose SDK got rejected in #883) — not the same category of concern. And Hermes specifically isn't a native OpenShell agent like Claude Code/Codex/Copilot/OpenCode (which work out of the box) — it needs the additional NemoClaw layer, so that's three stacked dependencies for this particular runtime choice, not one.

Given the spike and ADR already exist, I'd suggest reading through `SPIKE-OPENSHELL-OAS-RUNTIME.md`, `ADR-KA-002`, and #1535 first — they'll save you from re-deriving ground that's already covered, and point directly at the gaps (§8 coexistence, and the audit relay itself) that are actually still open and would benefit from a follow-up spike.

Issue #1435, opened 2026-06-16
AF: Status event text truncated mid-sentence in SSE stream (reasoning events)


jordigilh:
## Summary

Agent reasoning text in `status-update` events (`metadata.type: "reasoning"`) is truncated mid-sentence at ~200 characters before reaching the Console SSE stream. The Console renders it verbatim, making the agent appear unreliable to operators.

**Parent issue**: #1434 (tool result truncation — fixed by index-based prioritization)
**Related**: #1433 (event channel buffer — fixed by PrioritySend)

## Observed Behavior

From #1434 comment (2026-06-15):

```
The target resource appears to be a Deployment called web-frontend in demo-webui namespace. Let me use kubernaut_investigate_alert with:

alert_name: KubePodCrashLooping
api_version: apps/v1
k.....
```

The last line should read `kind: Deployment` but is cut off at ~200 chars. This is a `status-update` event arriving already truncated over the SSE stream.

## Root Cause Hypothesis

The tool result trimmer (`trimEventFunctionResponses` in `pkg/apifrontend/session/trimming.go`) only operates on `FunctionResponse` parts — it does not explain `TextPart` truncation in status events. A per-message or per-event text size limit is applied somewhere between ADK session event emission and the SSE serializer.

Likely locations:
1. ADK library's event serialization (text part size limits)
2. SSE frame serializer (per-frame byte cap)
3. A2A `TaskStatusUpdateEvent` text length constraints
4. Streaming executor event emission layer

## Secondary Issue: Re-invocation after context cancellation

The re-invocation loop in `streaming_executor.go` does not check `ctx.Err()` before re-invoking. When the context is cancelled (e.g., client SSE disconnect), it still attempts up to `MaxReinvocations = 3` re-invocations, all failing immediately with `context canceled`:

```
23:39:05 re-invoking agent after text-only turn end (reinvoke_count: 1)
23:39:05 a2a task execution failed: context canceled (reinvoke_count: 1)
23:39:05 re-invoking agent after text-only turn end (reinvoke_count: 2)
...
```

`NeedsReinvocation` should check context cancellation, or the executor loop should break on `context.Canceled`/`context.DeadlineExceeded`.

## Impact

- **UX**: Operators see garbled partial reasoning text during live investigations
- **Trust**: Truncated output makes the agent appear unreliable
- **Token waste** (secondary): Ghost re-invocations from truncation-triggered text-only turns

## Environment

- Dev cluster, observed 2026-06-15
[...]

Issue #1776, opened 2026-07-29
BR-SESS-013 reinvocation loop fails against the real a2a-go executor: text-only turn end always errors with "consumer stopped"


jordigilh:
## Summary

While building E2E-FLEET-016 (real AF-via-A2A fleet `kubectl_get`, for #1768 Track 1), discovered that AF's `BR-SESS-013` reinvocation loop is structurally broken against the real `a2a-go` executor. It has never been exercised by any integration or E2E test — every existing test mocks `s.inner`, so this defect has zero prior coverage.

## Root cause

`StreamingExecutor.runReinvocationLoop` (`pkg/apifrontend/launcher/streaming_executor.go:191`) implements BR-SESS-013: when a session is `Active` and the agent's last turn is plain text (no tool call), inject a synthetic "continue" message and call `s.inner.Execute(ctx, reqCtx, queue)` again:

```go
err := s.inner.Execute(ctx, reqCtx, queue)
if err == nil && s.adkSessionSvc != nil {
    err = s.runReinvocationLoop(ctx, reqCtx, queue, username)  // calls s.inner.Execute again, same ctx+queue
}
```

`s.inner` is the real `a2a-go` `a2asrv` executor in production. Its internal `runProducerConsumer` (`github.com/a2aproject/a2a-go@v0.3.15/internal/taskexec/execution_handler.go:84`) pairs exactly one producer goroutine (the agent loop) with one consumer goroutine (which watches the event queue and returns the final `SendMessageResult`) **per `Execute()` call**, against the queue passed in. Once the first call's consumer sees the text-only turn as a complete/final result and returns, the shared errgroup context is cancelled with a sentinel cause:

```go
errConsumerStopped := errors.New("consumer stopped")  // execution_handler.go:125
```

`runReinvocationLoop`'s second `s.inner.Execute()` call reuses the *same* `queue` and derives from the *same* `ctx` — but that queue's consumer has already returned, and the shared context is already on its way to cancellation. The reinvocation's model call is aborted almost immediately:

```
"msg":"a2a task execution failed","error":"failed to call model: doRequest: error sending request: Post \"[URL_REDACTED]\": consumer stopped"
```

## Reproduction (100% reproducible, 2/2 live runs)

- `test/e2e/fleet/16_af_real_fleet_investigation_test.go` (E2E-FLEET-016, branch `test/1768-af-fleet-e2e-coverage`)
[...]

PR #1780, opened 2026-07-30
fix(apifrontend): gate a2a task-completion event during BR-SESS-013 re-invocation (#1774)

### LINKED ISSUE #1774: Long A2A investigation streams still get 'context canceled' mid-stream (~200-350s) after ConsoleRoute HAProxy timeout fix
## Summary

Long-running A2A investigation streams from the console are being cut off by a `context canceled` error at a fairly consistent ~197-352s mark, even after confirming the operator-side HAProxy Route timeout fix (kubernaut-operator's `ConsoleRoute()` now sets `haproxy.router.openshift.io/timeout: 3600s` — see [jordigilh/kubernaut-operator#268](https://github.com/jordigilh/kubernaut-operator/issues/268)) is deployed and active on the cluster. The stream cutoff recurred even with that fix live, so HAProxy's default 30s Route timeout is not the (or not the only) cause.

## Symptom

`apifrontend` logs an "a2a task execution failed" error wrapping `context canceled` partway through what would otherwise be a longer-running investigation (the exact elapsed time before cutoff has varied across occurrences: ~197s in one session, ~352s in another). The console UI shows the stream simply stop / disconnect mid-investigation.

## What's been ruled out

- **HAProxy Route timeout**: `ConsoleRoute()` correctly sets a 3600s HAProxy timeout annotation (confirmed deployed and applied on the test cluster); cutoff still occurred at ~352s in a session after this fix was live, so it is not the sole cause.
- **oauth2-proxy / nginx sidecar timeouts**: Reviewed `ConsoleNginxConfigMap` — the primary `/a2a/` location block sets `proxy_read_timeout 3600s`. (Note: a *second* nginx location block in the same ConfigMap still has a 30s `proxy_read_timeout` — worth double-checking this isn't matching the A2A path in some deployments, but it did not appear to be the active path in the traced session.)
- **Prometheus/Thanos-querier auth or connectivity**: confirmed via direct `curl` from a debug pod with a valid token — the querier responds immediately (no hang, no timeout at that hop).
- **Client-side network/disconnect**: `context.Canceled` (as opposed to `context.DeadlineExceeded`) implies explicit cancellation rather than a timeout firing, which points at some layer explicitly calling a cancel function or the browser/client explicitly closing the connection — but this hasn't been isolated to a specific hop yet.

## What's still open

The exact source of the cancellation at this elapsed-time boundary has not been pinpointed. Candidates not yet fully eliminated:
- `pkg/apifrontend/config/config.go`'s `DefaultLLMTimeoutSeconds = 120` and `pkg/apifrontend/launcher/model.go`'s `buildLLMHTTPClient` — confirmed this specific default does not apply to the Anthropic/VertexAI provider path, but worth re-checking whether some *other* timeout in the ADK/agent-launcher stack (not the raw LLM HTTP client) is derived from or related to this constant.
- Whether the ADK framework itself (google/adk-go, used by `apifrontend`'s A2A agent launcher) has its own internal execution/step timeout that's independent of HTTP-layer timeouts.
- Browser/client-side `EventSource`/fetch timeout or reconnect behavior — the "reconnect" pattern seen in one session's logs (duplicate boilerplate lines, `kubernaut_reconnect` session-pool reuse) may or may not be related to this same root cause; not conclusively tied together yet.
- Whether `context.WithoutCancel` usage in `sse_disconnect_ctx.go` (`WithSSEDisconnectCtx`/`SSEDisconnectCtxFromContext`) has an edge case where the *original* request context still ends up propagated/cancelled into the detached execution goroutine under some condition, re-introducing the exact bug that mechanism was built to avoid.

## Ask

This needs an owner with more direct visibility into the ADK agent-launcher's internal lifecycle/timeout handling and browser-side SSE/EventSource behavior to instrument and pin down which hop is issuing the cancellation. Recommend adding explicit context-cancellation-reason logging at each hop boundary (nginx/oauth2-proxy access logs already reviewed for one window; the launcher and ADK layers are the main remaining blind spots) so the next occurrence can be traced precisely instead of inferred.

jordigilh:
## Summary

Fixes #1774 — long-running investigation streams intermittently failed with `context canceled` errors 194–352s in, with no consistent trigger. Root cause traced through `a2a-go` and ADK internals (full analysis in [DD-AF-009](docs/architecture/decisions/DD-AF-009-completion-gate-reinvocation.md)):

- `a2a-go`'s task processor cancels the shared execution context the instant it observes the **first** `TaskStatusUpdateEvent{State: Completed}` — it assumes the task is over and stops draining the queue.
- ADK emits that event at the end of **every** agent turn, including a text-only turn with no tool call.
- BR-SESS-013's re-invocation loop then calls `Execute` again on that same (already-cancelled) context, believing the task is still active. The failure surfaces later, whenever the re-invoked call's next outbound LLM/tool call happens to check `ctx.Err()` — explaining the 194–352s variance.
- This is a different failure mode than #1435, which only guards against *re-invoking again* once cancellation is already visible; it does nothing for the in-flight race between the premature Completed event and the reinvocation decision itself.

### Fix

Wrap the queue handed to the inner A2A executor in a `completionGateQueue` (`streaming_executor.go`) for the lifetime of one request:

- Non-terminal events (Working status, artifacts, reasoning deltas) are forwarded immediately — zero change to live streaming visibility, on every turn including reinvoked ones.
- A `Completed` status is **held**, not forwarded.
- Before each re-invocation, the held event is **dropped** — it was premature.
- Once the re-invocation loop is confirmed done, the held event (the genuinely final one) is **flushed** — only then does `a2a-go` see task completion.

Two architectural options were spiked (pushing the fix into the ADK `Runner` internals vs. gating at the AF-owned queue boundary). Went with the queue-boundary gate: contained entirely in AF's own code, depends only on the stable public `a2a.TaskState` contract rather than undocumented ADK internals, and is straightforward to unit test in isolation. Details and rejected alternatives in DD-AF-009.

## Test plan

- [x] RED: added `UT-AF-1774-001..004` to `streaming_executor_test.go`; confirmed 2 of 4 fail against the pre-fix code (4 Completed events reach the queue instead of 1; a stale Completed event survives a failed re-invocation)
- [x] GREEN: full `launcher` suite passes (280/280 specs, 0 regressions)
- [x] REFACTOR: extracted shared `writeWorkingThenCompleted` test helper to remove duplication between the two new test doubles; re-verified build/tests/lint after
- [x] `go build ./...` clean
- [x] `golangci-lint run` — 0 issues on `pkg/apifrontend/launcher/...`
- [x] Full `pkg/apifrontend/...` suite passes — 0 regressions
- [x] Design decision documented: [DD-AF-009](docs/architecture/decisions/DD-AF-009-completion-gate-reinvocation.md)

No E2E coverage added — this fix is purely internal to `StreamingExecutor`'s handling of an already-wired component; the existing `IT-AF-REINV-W01`/`W01b` wiring tests (BR-SESS-013) continue to pass unchanged, confirming no wiring regression.

Backport to `main` to follow once this lands on `release/v1.5`.

### DIFF
--- a/.github/actions/install-podman/action.yml
+++ b/.github/actions/install-podman/action.yml
@@ -1,0 +1,52 @@
+name: Install Podman (crun-pinned)
+description: >
+  Installs podman via apt, then replaces the runner's bundled crun binary
+  with a pinned release known to be compatible with podman's generated OCI
+  Runtime Spec configs (Issue #1769). GitHub's ubuntu-24.04 runner images
+  ship a crun version older than 1.14.3, which rejects podman-generated
+  container configs with "Error: OCI runtime error: crun: unknown version
+  specified" -- a deterministic version-compatibility bug, not a transient
+  flake: it reproduces on every `podman run`/`kind create cluster` call
+  that hits it, on that exact podman+crun pairing, until crun is upgraded.
+  See https://github.com/containers/podman/issues/27272 (root cause: crun
+  rejects OCI Runtime Spec v1.2.1 configs below v1.14.3) and
+  https://github.com/containers/crun/releases/tag/1.14.3 (fix version).
+
+inputs:
+  crun-version:
+    description: Pinned crun release to install (must be >= 1.14.3).
+    required: false
+    default: '1.28'
+
+runs:
+  using: composite
+  steps:
+    - name: Install podman
+      shell: bash
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y podman
+
+    - name: Pin crun >= 1.14.3 (Issue #1769)
+      shell: bash
+      run: |
+        CRUN_VERSION="${{ inputs.crun-version }}"
+        ARCH=$(dpkg --print-architecture)
+        case "$ARCH" in
+          amd64) CRUN_ARCH="linux-amd64" ;;
+          arm64) CRUN_ARCH="linux-arm64" ;;
+          *) echo "Unsupported architecture for crun pin: $ARCH"; exit 1 ;;
+        esac
+
+        echo "Runner-provided crun: $(crun --version 2>&1 | head -1 || echo 'not found')"
+
+        curl -fsSL -o /tmp/crun \
+          "https://github.com/containers/crun/releases/download/${CRUN_VERSION}/crun-${CRUN_VERSION}-${CRUN_ARCH}"
+        chmod +x /tmp/crun
+        sudo mv /tmp/crun /usr/bin/crun
+
+        echo "Pinned crun: $(crun --version | head -1)"
+
+    - name: Print podman version
+      shell: bash
+      run: podman --version
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -506,6 +506,17 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y podman
+      # crun pin (#1784): the ubuntu-latest runner image's preinstalled crun
+      # predates 1.14.3 and cannot parse the OCI Runtime Spec v1.2.1
+      # config.json that current Podman generates, failing every container
+      # start with "OCI runtime error: crun: unknown version specified"
+      # (https://github.com/containers/podman/issues/27272, resolved in
+      # crun v1.14.3). apt's crun lags behind apt's podman, so pin explicitly.
+      - name: Pin crun >= 1.14.3 (podman/crun OCI spec compat)
+        run: |
+          sudo curl -fL "https://github.com/containers/crun/releases/download/1.14.3/crun-1.14.3-linux-amd64" -o /usr/bin/crun
+          sudo chmod +x /usr/bin/crun
+          crun --version
       - name: Setup envtest (K8s test binaries)
         run: |
           go install sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.23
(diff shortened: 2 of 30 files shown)


jordigilh:
Closing as superseded by #1783 (backport of main's #1776 fix).

#1774 and #1776 root-cause to the identical mechanism: `a2a-go`'s `internal/taskexec/execution_handler.go` cancels the shared execution context the instant its consumer observes the first `TaskStatusUpdateEvent{State: Completed}`, while BR-SESS-013's re-invocation loop calls `Execute` again on that same (soon-to-be-canceled) context.

This PR's `completionGateQueue` fixes that symptom by gating/holding the Completed event at the AF-owned queue boundary. #1783 (matching `main`'s architecture) fixes the same root cause more fundamentally: it moves the re-invocation loop *inside* `reinvokingRunner.Run()`, so `adka2a.Executor.process()` only ever calls `writeFinalTaskStatus` once per request — after every re-invocation cycle has finished — rather than once per re-invocation. Verified directly in the vendored ADK source (`adk@v1.5.1/server/adka2a/v2/executor.go`): `process()` loops over a single `Runner.Run()` call and calls `writeFinalTaskStatus` only once that loop is exhausted; with `reinvokingRunner` wired in, there is no remaining code path where more than one Completed event could ever be written.

Since `main` only has the `reinvokingRunner` approach (no `completionGateQueue` anywhere), and the project's standing principle is that `release/v1.5` and `main` should only differ in unavoidable implementation details — not in architectural approach — #1783 is the one to carry forward. This PR's diff is also built against the old `WithReinvocation`/`runReinvocationLoop` structure that #1783 removes entirely, so it would no longer apply cleanly once #1783 merges.

DD-AF-009's analysis and test coverage remain valuable references; the design decision doc can be updated to point at the `reinvokingRunner` mechanism if useful, but the code change here is not needed.

Issue #1795, opened 2026-07-31
AF->KA MCP streaming investigate call hits consistent 60s dial timeout (ConnectTimeout ignored for TLS transport)


jordigilh:
## Summary

AF's `kubernaut_investigate` MCP tool call to KubernautAgent (KA) over the streaming/interactive path (`StartInvestigation` -> dedicated MCP session via `StreamableClientTransport`) consistently fails with a **~60 second** TCP dial timeout to KA's ClusterIP:8443, even though KA is healthy, reachable, and NetworkPolicy explicitly allows AF->KA traffic on that port. The equivalent non-streaming path (`message/send`, `E2E-FP-1189-003`) does not exhibit this failure.

## Discovery

Found while building `E2E-FP-1189-005` (Console-equivalent SSE interactive remediation test, full RO->SP->AA->KA->WE pipeline). Confirmed reproducible on a clean environment (re-verified after ruling out host-hibernation-induced clock/network skew as a contributing factor -- the podman VM's clock was independently found ~24min skewed after a host hibernate cycle and corrected, but the bug reproduced identically afterward).

### Evidence (AF logs, `apifrontend/agent/root.go` + `tools/ka_investigate_mcp.go`)

```
01:10:10.708Z  tool call started            tool=kubernaut_investigate callID=call_1fcec8e88dce
01:10:11.281Z  StartInvestigation: calling MCP client   rr_id=rr-... ka_session_id=... ctx_err=null
01:11:11.283Z  tool call failed             error="start MCP investigation: MCP connect for investigation:
               calling \"initialize\": sending \"initialize\": rejected by transport:
               Post \"https://kubernaut-agent.kubernaut-system.svc:8443/api/v1/mcp/\":
               dial tcp 10.96.193.11:8443: i/o timeout"
```

Exactly **60.002s** between the connect attempt and the failure -- reproduced across multiple independent runs (with and without host hibernation in between), always landing on this same duration.

Meanwhile KA's own log shows the session was created successfully at `01:10:11.224Z` (`session/manager_interactive.go:78`, "interactive session created with context (pending)") and then falls completely silent for the entire 60s window -- no further log lines until after AF has already given up. This suggests the TCP-level connection from AF genuinely never completes (not merely slow on KA's application side).

[...]

Issue #1794, opened 2026-07-30
KA live SSE 'complete' event never carries RCA data (only the reconnect path does)


jordigilh:
## Summary

The live investigation SSE stream's terminal `complete` event never carries the investigation's structured RCA result (`InvestigationResult` — severity, confidence, causal chain, RCA summary, etc.), even though that data exists and is available moments later. A client that reconnects *after* completion gets the full result; a client that was streaming live throughout gets an empty `complete` event with no `Data`.

## Discovery

Found while building `E2E-FP-1189-005` (Console-equivalent SSE interactive remediation test, full RO→SP→AA→KA→WE pipeline). The test asserts that AF's `kubernaut_investigate` bridge renders an RCA to the user during the live SSE stream (either a `metadata.schema=early_rca` status-update via `emitEarlyRCA`, or a `metadata.schema=investigation_summary` fallback artifact via `emitFallbackInvestigationArtifact` — see `pkg/apifrontend/tools/ka_investigate_bridge.go`). Neither ever appears, despite the investigation completing successfully with 17 status frames and 4 artifact frames observed on the wire.

## Root Cause

In `internal/kubernautagent/session/manager.go`, `runInvestigation` emits the completion event **before** the session's `Result` is persisted:

```go
result, fnErr := fn(bgCtx)
m.emitCompleteEvent(id)                 // <-- fires with empty Data; sess.Result is still nil here
if fnErr != nil {
    m.handleInvestigationFailure(bgCtx, id, correlationID, result, fnErr)  // sets sess.Result via store.Update, AFTER the event already went out
    return
}
m.handleInvestigationSuccess(bgCtx, id, correlationID, result)             // same ordering issue
```

`emitCompleteEvent` (`internal/kubernautagent/session/manager_events.go`) always sends:

```go
case ch <- InvestigationEvent{Type: EventTypeComplete}:
```

with no `Data` field — it never receives `result` as a parameter, so it has nothing to marshal even if it wanted to.

Contrast this with the **reconnect/late-join path**, `terminalSessionSSE` (`internal/kubernautagent/server/handler.go`), which correctly marshals `sess.Result` into the complete event's `Data` — because by the time a client reconnects, `Store.Update()` has already run and `sess.Result` is populated.

## Impact

[...]

PR #1800, opened 2026-07-31
fix(1794-backport-v1.5): live SSE complete event never carried RCA data

jordigilh:
## Summary

Backports the #1794 fix from `main`/v1.6 to `release/v1.5`.

- `Manager.emitCompleteEvent` fired the live SSE `EventTypeComplete` event before the investigation result was available, so a live-streaming Console user driving an interactive investigation via SSE never saw the RCA render on completion — only a reconnecting client (`terminalSessionSSE`) got it, since that path reads the already-persisted `sess.Result`.
- Also fixes a related SI-10 data leak in `terminalSessionSSE`, which raw-marshaled the full `session.Result` (including internal-only fields like `workflow_id`) instead of the bounded RCA subset.

Both sites now call the pre-existing `MarshalRCASubset` helper (`rca_event_payload.go`, unchanged) — no new types/components introduced.

**#1795 (AF→KA NetworkPolicy egress gap)**, found alongside #1794 while auditing E2E coverage on `main`, does **not** apply here — `apifrontend`'s Helm chart on `release/v1.5` has no `NetworkPolicy` resource at all yet, so there is nothing to backport for it.

## Test plan

- [x] `go vet ./internal/kubernautagent/...` — clean
- [x] `golangci-lint run ./internal/kubernautagent/session/... ./internal/kubernautagent/server/...` — 0 issues
- [x] New tests added following TDD, RED confirmed before each fix, GREEN after:
  - `UT-KA-1794-001/002` (`session` package, live complete event)
  - `UT-KA-1794-003` (`server` package, terminal/reconnect complete event)
- [x] `go test ./internal/kubernautagent/...` — all packages pass

### DIFF
--- a/internal/kubernautagent/server/handler.go
+++ b/internal/kubernautagent/server/handler.go
@@ -531,11 +531,14 @@ func (h *Handler) terminalSessionSSE(sessionID string) (agentclient.SessionStrea
 		Type:  session.EventTypeComplete,
 		Turn:  0,
 		Phase: "complete",
+		// #1794: use the bounded RCA subset (severity/confidence/rca_summary/...)
+		// rather than a raw dump of sess.Result, which also carries internal
+		// workflow/validation state (workflow_id, validation_attempts_history,
+		// etc.) that must not leak to AF/Console (SI-10). Matches what the live
+		// path (Manager.emitCompleteEvent) now attaches.
 	}
 	if sess.Result != nil {
-		if data, marshalErr := json.Marshal(sess.Result); marshalErr == nil {
-			completeEvent.Data = data
-		}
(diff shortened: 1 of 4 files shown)


jordigilh:
Follow-up on the PR description's #1795 note: verified against \`kubernaut-operator\` (not just this chart) that the AF->KA NetworkPolicy egress gap genuinely doesn't apply there either -- the operator grants AF a broad \`app.kubernetes.io/part-of: kubernaut\` egress rule on 8443 rather than per-service selectors, so it was never missing \`kubernaut-agent\` the way this chart's \`apifrontend/networkpolicy.yaml\` was. Full detail: https://github.com/jordigilh/kubernaut/issues/1795#issuecomment-5143092805. No action needed for the operator.

Issue #1408, opened 2026-06-12
Contract: AF must emit structured investigation_summary artifact (no text RCA)


jordigilh:
## Problem

The console expects the AF to emit a **structured `investigation_summary` artifact** (DataPart) when investigation completes. Instead, the AF currently:

1. Streams the RCA as **plain text** via status-update or text artifact parts
2. May or may not follow up with a structured DataPart
3. Sometimes hangs during workflow discovery and never completes

This causes:
- RCA renders as raw markdown instead of the designed `RCACard` component (severity bar, badge, causal chain, metadata)
- Double/triple rendering: text RCA → replaced by another text → replaced by summary-only card
- "Stuck" state when workflow discovery fails — user sees text and a spinner forever

Screenshots from live dev deployment (v0.5.0):
- Text RCA in agent bubble (wrong) vs RCACard with workflows (correct expected behavior)

### Phase 1: Investigation (streaming)

AF emits **status-update events** during investigation. Console renders these in the ThinkingPanel (collapsible, shows reasoning in real-time).

```json
{
  "kind": "status-update",
  "status": { "state": "working", "message": { "role": "agent", "parts": [{"kind": "text", "text": "Checking pod logs..."}] } },
  "metadata": { "type": "reasoning" }
}
```

Supported `metadata.type` values for thinking: `"reasoning"`, `"status"`, `"investigation"`, `"preflight"`, `"tool_call"`

**UX result:** User sees animated ThinkingPanel with live reasoning steps. Header shows "Investigating" with pulsing green dot.

### Phase 2: Investigation Complete → Structured Artifact

When investigation AND workflow discovery complete, AF emits a **single `artifact-update`** event with a `DataPart`:

```json
{
  "kind": "artifact-update",
  "taskId": "...",
  "contextId": "...",
  "artifact": {
    "artifactId": "investigation-result-001",
    "metadata": {
      "type": "investigation_summary",
      "schema": "investigation_summary",
      "schema_version": "1.0"
    },
    "parts": [
      {
        "kind": "data",
        "mediaType": "application/json",
        "data": {
          "type": "investigation_summary",
          "summary": "Pod web-frontend is crash looping due to an invalid directive in app-config ConfigMap, introduced by an ArgoCD GitOps sync.",
          "rca": {
            "severity": "critical",
            "confidence": 0.95,
            "target": "Deployment/web-frontend (demo-webui)",
            "causal_chain": [
[...]

Issue #1302, opened 2026-05-27
AF agent stalls silently after tool failure — no SSE output, eventual network error


jordigilh:
## Description

After a tool call fails (e.g. `kubectl_list` with `cannot resolve GVK for kind "pods"`), the AF agent stops producing any output on the SSE stream. The stream remains open but no events are emitted — no status updates, no error messages, no LLM responses. Eventually the client-side times out and shows "Error: network error".

## Observed behavior

1. A2A stream opens successfully
2. Agent starts investigation, calls `kubectl_list`
3. Tool fails: `invalid input: cannot resolve GVK for kind "pods"`
4. **No further log output from AF** — complete silence
5. Client-side SSE eventually times out → "Error: network error"

## Expected behavior

After a tool failure, the agent should:
- Report the error to the user via the SSE stream
- Either retry with corrected input or gracefully fail the task
- Not stall indefinitely with no output

## Relevant logs

```
14:04:16 a2a stream opened  task_id=019e69c0-166e-7adb-aa28-d0f2cb46ac06
14:04:16 session created
14:04:28 [audit-tool-callback] tool="kubectl_list" outcome=failure error=invalid input: cannot resolve GVK for kind "pods"
(... no further output ...)
```

No stream close, no error, no task state change — just silence.

## Environment

- kubernaut `feat/1282-af-agent-quality` branch, commit `fe2bd42`
- ADK v1.2.0, a2a-go v0.3.15
- Vertex AI with claude-sonnet-4-6

jordigilh:
## Still reproducing on a94d67d

AF agent still stalls silently after tool failures. Sequence observed:

1. A2A stream opens successfully (`a2a stream opened`)
2. AF calls `kubernaut_discover_workflows` → fails with `not_driving: You must send action=takeover before sending messages`
3. AF calls `kubernaut_stream_investigation` → fails with `not found: the requested investigation session does not exist`
4. AF calls `kubernaut_discover_workflows` again → same `not_driving` error
5. **No more AF log output** — the agent produces no further SSE events
6. Client times out and shows "Error: network error"

The AF agent is not recovering from tool failures and not producing any output to the SSE stream, causing the client to hang until timeout.

Image: `ghcr.io/jordigilh/kubernaut/apifrontend@sha256:fdd7e49ca691a2f4939a452c21840505517607f6ccacb0d4b47f097452cbb8c6`

jordigilh:
[...]

PR #1830, opened 2026-08-01
fix(apifrontend): stop leaking raw tool-call args when alert tools are unavailable (#1658)

### LINKED ISSUE #1658: AF: raw kubectl_list tool-call args leak through as chat response when list_alerts is unavailable
## Summary

When `severityTriage.enabled: false` (i.e. `spec.monitoring.enabled: false` on the Kubernaut CR), API Frontend does not register the `list_alerts`/`get_alert_details`/`kubernaut_investigate_alert` tools (gated in `root.go` on `cfg.PromClient != nil`). If a user then asks the assistant something like "list active alerts", the LLM has no purpose-built alert tool available and falls back to the generic `kubectl_list` tool, guessing `kind: "Alert"` — which is not a real Kubernetes/Kubernaut resource kind (alerts live in Prometheus, not as a CRD).

`kubectl_list`'s RESTMapper lookup fails to resolve `Alert` as a kind (`ErrInvalidInput`), and instead of the assistant explaining the failure or clarifying that alert querying isn't configured, the raw tool-call arguments are surfaced back to the user as the final chat response:

```json
{"kind": "Alert", "namespace": ""}
```

This reproduced in an airgapped OCP 4.18 + kagenti deployment where `spec.monitoring.enabled: false` was explicitly set on the Kubernaut CR (disabling the Prometheus/Thanos-querier wiring for AF).

## Root cause

- `pkg/apifrontend/agent/root.go`: `list_alerts`, `get_alert_details`, `kubernaut_investigate_alert` are only added to the tool list `if cfg.PromClient != nil`.
- `pkg/apifrontend/tools/kubectl_list.go`: `KubectlListArgs{Kind, Namespace, LabelSelector}` — the exact shape that ends up rendered to the user, confirming the LLM invoked `kubectl_list` with a guessed `kind`.
- The system prompt (`pkg/apifrontend/agent/prompt.go`) lists `list_alerts`/`get_alert_details` alongside `kubectl_get`/`kubectl_list` for "observation" without conditioning the guidance on whether alert tools are actually registered for this deployment, so the model has no signal that alert querying is unavailable and picks the closest tool instead of saying so.
- Whatever surfaces the model's final answer (or the tool-call error) to chat is echoing the tool-call arguments verbatim rather than a natural-language explanation of the failure — this is the part that most needs fixing, since it will resurface for *any* tool the model misuses this way, not just alerts.

## Suggested fixes (any combination)

1. **Prompt fix (cheap, immediate):** only mention `list_alerts`/`get_alert_details` in the system prompt when `cfg.PromClient != nil`; otherwise explicitly instruct the model that alert querying is unavailable and it should say so rather than guess a `kubectl_list` kind.
2. **Tool-call error handling:** when a tool call errors (e.g. `ErrInvalidInput` from an unresolvable kind), ensure the model always produces a natural-language final answer describing the failure — never let raw tool-call args/errors leak through as the visible response. This looks like it could affect any tool, not just `kubectl_list`.
3. Consider a `kubectl_list`/`kubectl_get` guard that returns a friendlier error for well-known "looks like it should exist but doesn't" kinds (e.g. `Alert`, `Pod` typos) pointing the model at the correct tool.

## Related

- `spec.monitoring.enabled` (operator CRD field) gating this behavior is currently undocumented outside of a bare Go struct comment (`api/v1alpha1/kubernaut_types.go`) — no prose docs mention that disabling monitoring silently removes alert tools from AF. Documented as a stopgap in the [air-gap install gist troubleshooting section](https://gist.github.com/jordigilh/687547d5d8a792add5476793486d3aa1#consolechat-returns-an-empty-kind-alert-namespace--for-list-active-alerts); consider adding this to in-repo docs as well.

jordigilh:
## Summary

Fixes #1658. When `severityTriage.enabled: false` (`cfg.PromClient == nil`), AF does not register `list_alerts`/`get_alert_details`/`kubernaut_investigate_alert`, but the system prompt advertised them unconditionally. The LLM would then fall back to `kubectl_list` with a guessed, invalid `kind: "Alert"` (Alert is not a Kubernetes resource), and instead of explaining the failure, echo the raw failed tool-call arguments back to the user as the final chat response.

Root cause investigation confirmed:
- Kubernaut's own tool-error plumbing (`toolErrorPart`, #1302) and the ADK agent loop already correctly route tool errors back to the LLM (never bypassing it to render raw args) — the leak was the LLM's own next-turn text, produced with no explicit guardrail, echoing the failed call.
- Two prompt gaps caused this: (1) alert tools were advertised regardless of actual registration, and (2) the existing "explain tool failures via `present_decision`" mandate (#1408) was scoped only to `kubernaut_investigate`/`kubernaut_discover_workflows`, not generic observation tools like `kubectl_list`.

## Changes

- `pkg/apifrontend/agent/prompt.go`: `BuildInstruction`/`NewInstructionProvider` take a new `alertToolsEnabled bool`, threaded from the same `cfg.PromClient != nil` gate already used to register the tools (`root.go`'s `buildToolList`). When disabled, the prompt omits alert-tool mentions and explicitly states alert querying is unavailable, instructing the model not to guess a `kubectl_list`/`kubectl_get` kind for alerts.
- `pkg/apifrontend/agent/prompt.txt`: added a general behavioral constraint — any tool error must be explained in natural language; raw tool-call arguments/JSON must never be surfaced verbatim. This generalizes beyond the two investigation tools already covered by the `present_decision` mandate.
- `cmd/apifrontend/mcp_a2a_handlers.go`: wires `alertToolsEnabled := d.Backends.PromClient != nil` into both prompt-building calls.
- `pkg/apifrontend/tools/kubectl_list_test.go`: closes a coverage gap — `kubectl_list` had no test for an unresolvable/guessed kind (`kubectl_get` already had `UT-AF-1230-009`); added `UT-AF-1658-020` using `Kind: "Alert"`, the exact repro from the issue.
- `docs/testing/1658/TEST_PLAN.md`: full test plan with BR mapping (BR-AI-023 hallucination/invalid-response handling, BR-AI-024 graceful degradation) and pyramid invariant inventory.

Deliberately **not** implemented (per the issue's "any combination" framing): a `kubectl_list`/`kubectl_get` runtime guard special-casing well-known bad kind guesses like `"Alert"`. The existing `cannot resolve GVK for kind "Alert"` error, combined with the two prompt fixes above, already gives the model enough signal without adding kind-name special-casing.

## Test plan

- [x] `UT-AF-1658-001..005, 010` (new, `pkg/apifrontend/agent/prompt_test.go`) — prompt content conditioning + generic tool-error mandate
- [x] `UT-AF-1658-020` (new, `pkg/apifrontend/tools/kubectl_list_test.go`) — unresolvable kind coverage gap
- [x] Full `pkg/apifrontend/agent` suite (167 specs) — PASS, 0 regressions
- [x] Full `pkg/apifrontend/tools` suite (632 specs) — PASS, 0 regressions
- [x] Full `pkg/apifrontend/...` suite — PASS
- [x] `go build ./...`, `go vet ./...` — PASS
- [x] `golangci-lint run` on changed packages — 0 issues

🤖 Generated with [Cursor](https://cursor.com)

### DIFF
--- a/cmd/apifrontend/mcp_a2a_handlers.go
+++ b/cmd/apifrontend/mcp_a2a_handlers.go
@@ -137,9 +137,10 @@ func sessionServiceForAgent(d *handlerDeps) *session.CRDSessionService {
 
 // buildRootAgentConfig assembles the AgentConfig passed to agentpkg.NewRootAgent.
 func buildRootAgentConfig(d *handlerDeps, llmModel model.LLM, sessionSvcForAgent *session.CRDSessionService) agentpkg.AgentConfig {
+	alertToolsEnabled := d.Backends.PromClient != nil
 	return agentpkg.AgentConfig{
-		Instruction:           agentpkg.BuildInstruction(d.Cfg.Session.Namespace),
-		InstructionProvider:   agentpkg.NewInstructionProvider(d.Cfg.Session.Namespace),
+		Instruction:           agentpkg.BuildInstruction(d.Cfg.Session.Namespace, alertToolsEnabled),
+		InstructionProvider:   agentpkg.NewInstructionProvider(d.Cfg.Session.Namespace, alertToolsEnabled),
 		LLMModel:              llmModel,
 		Namespace:             d.Cfg.Session.Namespace,
 		K8sClient:             d.Backends.K8sClient(),
--- a/docs/testing/1658/TEST_PLAN.md
+++ b/docs/testing/1658/TEST_PLAN.md
@@ -1,0 +1,114 @@
+# Test Plan: AF Raw Tool-Call Arg Leak on Unavailable Alert Tools
+
+**Issue**: [#1658](https://github.com/jordigilh/kubernaut/issues/1658)
+**Business Requirements**: BR-AI-023 (detect/handle AI hallucinations or invalid responses), BR-AI-024 (graceful degradation when a dependency is unavailable)
+**Branch**: `fix/1658-af-tool-error-leak`
+**Created**: 2026-08-01
+**Status**: Complete — all scenarios in Section 3 pass
+
+---
+
+## 1. Purpose
+
+When `severityTriage.enabled: false` (`cfg.PromClient == nil`), API Frontend does not
+register `list_alerts`/`get_alert_details`/`kubernaut_investigate_alert` (gated in
+`pkg/apifrontend/agent/root.go`'s `buildToolList` on `cfg.PromClient != nil`). The system
+prompt (`pkg/apifrontend/agent/prompt.go`'s `BuildInstruction`), however, unconditionally
+advertised these tools regardless of whether they were actually registered. If a user then
+asked to "list active alerts", the LLM had no purpose-built alert tool available in its
+function-calling schema and fell back to the generic `kubectl_list` tool, guessing
+`kind: "Alert"` (not a real Kubernetes/Kubernaut resource kind). The tool call failed
+(`ErrInvalidInput: cannot resolve GVK for kind "Alert"`), and instead of explaining the
+failure in natural language, the model's next-turn text echoed the raw tool-call arguments
+back to the user as the final chat response.
+
+### Root cause (confirmed via code read, not just issue inference)
+
+Two independent gaps, both required to fully close this class of bug:
+
+1. **Prompt/registration mismatch** (`pkg/apifrontend/agent/prompt.go`): `BuildInstruction`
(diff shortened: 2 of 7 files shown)


jordigilh:
Backport for `release/v1.5` already opened: #1836 (ported independently to v1.5's pre-fleet-federation architecture, not a cherry-pick — see that PR's description for the two adaptation points). Not gated on this PR merging first, but should be reviewed/merged in the same spirit; if this PR's review changes the fix's approach, #1836 should be revisited to match before it merges.

PR #1836, opened 2026-08-01
fix(apifrontend): backport #1658 -- prevent raw tool-call arg leak when alert tools are disabled (v1.5)

jordigilh:
## Summary

Backport of the `main` fix ([PR #1830](https://github.com/jordigilh/kubernaut/pull/1830), not yet merged) to `release/v1.5`. Same root cause and fix on both branches: when `severityTriage.enabled: false` (`cfg.PromClient == nil`), the system prompt unconditionally advertised `list_alerts`/`get_alert_details`/`kubernaut_investigate_alert` even though they were never registered, so the LLM would fall back to guessing an invalid `kubectl_list` kind (e.g. `"Alert"`) and leak the raw, failed tool-call arguments to the user.

- `BuildInstruction`/`NewInstructionProvider` now take `alertToolsEnabled`, gated on the same `cfg.PromClient != nil` check used to register the tools, and omit/replace the alert-tool guidance in the prompt accordingly.
- `prompt.txt` gains a generic mandate: any tool-call error must be explained in natural language, never surfaced as raw arguments/JSON.
- `cmd/apifrontend/main.go` wires `alertToolsEnabled := deps.PromClient != nil` at the single inline `NewRootAgent` call site — v1.5 has no separate `mcp_a2a_handlers.go`, this wiring is inline in `main.go`, unlike `main`.
- `kubectl_list_test.go`'s new `UT-AF-1658-020` adapted to call `HandleKubectlList` with a raw `dynamic.Interface` (v1.5's actual signature) instead of `main`'s `DynamicResourceReader` wrapper, which doesn't exist on this branch (fleet federation is `main`-only).

## Dependency note

This PR is independent code (not a `git cherry-pick`) built by porting `main`'s diff onto `v1.5`'s pre-fleet-federation architecture. It does **not** depend on #1830 merging first, but the two should land with the same intent — if #1830's review surfaces changes to the fix's approach, this backport should be revisited to match.

## Test plan

- [x] `go build ./...`
- [x] `go vet` on all changed packages
- [x] `gofmt -l` clean
- [x] `golangci-lint run` — 0 issues
- [x] Full `pkg/apifrontend/agent` and `pkg/apifrontend/tools` unit suites pass, including new `UT-AF-1658-001..005`, `UT-AF-1658-010`, `UT-AF-1658-020`
- [x] `go vet` on `test/integration/apifrontend` (envtest-dependent, not run locally)

Refs #1658.

### DIFF
--- a/cmd/apifrontend/main.go
+++ b/cmd/apifrontend/main.go
@@ -945,9 +945,14 @@ func buildA2AHandler(ctx context.Context, cfg *config.Config, deps *backendDeps,
 
 	activeCtxRegistry := launcher.NewActiveContextRegistry(launcher.DefaultRegistryTTL, launcher.DefaultRegistryIdleTimeout)
 
+	// #1658: alertToolsEnabled must mirror the same cfg.PromClient != nil gate
+	// NewRootAgent uses to decide whether to register list_alerts/get_alert_details/
+	// kubernaut_investigate_alert, so the prompt never advertises tools that
+	// aren't actually callable.
+	alertToolsEnabled := deps.PromClient != nil
 	rootAgent, _, err := agentpkg.NewRootAgent(agentpkg.AgentConfig{
-		Instruction:           agentpkg.BuildInstruction(cfg.Session.Namespace),
-		InstructionProvider:   agentpkg.NewInstructionProvider(cfg.Session.Namespace),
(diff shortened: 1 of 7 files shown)


jordigilh:
Cross-referenced from #1830 (the `main` fix this backports). Linking here too for two-way traceability: if #1830's review changes the fix's approach before merging, this PR should be revisited to match.

PR #1841, opened 2026-08-01
fix(apifrontend): remove ungrounded LLM severity inference (Tier 3, #1839)

### LINKED ISSUE #1839: fix(apifrontend): remove ungrounded LLM severity inference (Tier 3) -- risks wrong workflow selection
## Summary

API Frontend's severity-triage pipeline (`pkg/apifrontend/severity/triage.go`) has a "Tier 3: Pure LLM Fallback" that classifies severity for a resource with **zero grounding evidence** -- no firing/pending Prometheus alert, no correlated Prometheus rule, nothing. It runs whenever `kubernaut_investigate`/`kubernaut_remediate` targets a resource that has no matching alert or rule (the common case for direct-resource investigation), which is any realistic deployment with Prometheus + an agent LLM configured (i.e. the default, not an edge case).

## Root cause

```135:184:pkg/apifrontend/severity/triage.go
// Tier 3: Pure LLM fallback
return t.runTier3(ctx, input)
```

`runTier3` calls `LLMTriager.TriagePure(ctx, input)`, whose prompt (`types.go`'s `BuildTriagePrompt`) gives the LLM only `namespace/kind/name/description` -- no alert, no rule, no confirming signal of any kind -- and asks it to classify severity as critical/high/warning/info. The LLM's answer is written into `RemediationRequest.Spec.Severity` with `SignalType: "alert"` hardcoded regardless of tier, making an invented severity indistinguishable at the CRD level from a real Alertmanager-sourced one (only `signalLabels["severity_source"]` differs).

That severity then filters KA's workflow catalog (`internal/kubernautagent/workflowcatalog/cache_filter.go`), so an LLM-guessed severity for a resource with no real signal can steer `kubernaut_select_workflow` toward the wrong remediation -- workflows are matched against labels defined by real, user-authored alert semantics, which an LLM has no grounds to reconstruct from a bare resource identity.

This was confirmed as intentional (not an oversight) via the original test plan (`docs/services/apifrontend/tests/92/test_plan.md`'s `BAC-T-06`: *"Pure LLM (Tier 3) used when no rules cover the target resource"*) and `NewTriager`'s explicit panic-if-nil-LLM design ("the pipeline requires an LLM fallback to guarantee a result") -- but the downstream workflow-selection risk was not accounted for when it was designed.

## Fix

Remove Tier 3 (`runTier3`, and `TriagePure` from the `LLMTriager` interface and all 4 implementations) entirely. Tier 1/1.5/2 (real alert/rule evidence) and Tier 2.5 (a real, label-correlated Prometheus rule exists, just not currently true) are unaffected -- there is something concrete to derive from in those cases. When Tier 1/1.5/2/2.5 all miss, `Triage()` now returns a clear error instead of guessing; `resolveCreateRRSeverity`/`HandleCreateRR` already abort RR creation on any `Triage()` error (existing code path, unchanged), and the LLM explains this to the user in natural language per #1658's prompt mandate rather than a `RemediationRequest` being silently created with a fabricated severity.

See DD-AF-010 for the full design record and alternatives considered (a default-off config toggle was considered and rejected in favor of outright removal).

## Scope note

This is a `main`-only fix for now; `release/v1.5` applicability will be assessed separately once this lands.

jordigilh:
## Summary

- Removes Tier 3 of the severity-triage pipeline (`runTier3`, and `TriagePure` from the `LLMTriager` interface and all four implementations: `GenAITriager`, `AnthropicTriager`, `OpenAICompatibleTriager`, `NoopLLMTriager`). Tier 3 asked the LLM to classify severity from a bare resource identity (namespace/kind/name/description) whenever no Prometheus alert or rule correlated to the target -- the common case for direct-resource investigation, not an edge case in any realistic deployment.
- That guess was written to `RemediationRequest.Spec.Severity` indistinguishably from a real Alertmanager signal (`SignalType: "alert"` hardcoded regardless of tier) and drove KA's workflow-catalog severity filter, risking the wrong remediation workflow for resources with no real signal behind them. An LLM has no grounds to reconstruct alert semantics that only exist in real, user-authored Prometheus rules.
- When Tier 1/1.5/2/2.5 all miss, `Triage()` now returns `severity.ErrSeverityUndetermined` and `HandleCreateRR` creates no RR, instead of fabricating a severity. This reuses the existing error-propagation path (`resolveCreateRRSeverity` already aborted RR creation on any `Triage()` error) and #1658's existing tool-error-to-natural-language prompt mandate to explain the failure to the user.
- Tier 2.5 (a real, label-correlated Prometheus rule exists, just not currently true) is unaffected -- there is still something concrete to derive from there.
- Updated all authoritative docs that described Tier 3 (`ARCHITECTURE.md`, `DATA_FLOW.md`, the `RB-AF-010` runbook, `DD-AF-001` cross-reference) so on-call/design docs don't describe a removed code path.
- Full alternatives-considered writeup in [DD-AF-010](docs/architecture/decisions/DD-AF-010-remove-ungrounded-severity-inference.md).

Closes #1839.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `golangci-lint run` -- 0 issues on all touched packages
- [x] `pkg/apifrontend/severity` -- 82/82 specs pass, including new `UT-AF-1839-001`/`UT-AF-1839-002` proving the fail-closed path and that the LLM is never invoked with zero grounding evidence
- [x] `pkg/apifrontend/tools` -- 631/631 specs pass, including new `UT-AF-1839-001` proving `HandleCreateRR` creates no RR and returns an explanatory error when severity can't be grounded
- [x] `cmd/apifrontend` -- passes (LLM-provider wiring unaffected, `LLMTriager` interface shrunk to one method)
- [x] Post-review pass caught and fixed a real regression: `IT-AF-1282-W04` relied on the removed Tier 3 to resolve a non-empty severity from zero evidence -- regrounded it with a real firing alert so it still proves its original wiring claim
- [x] Added `IT-AF-1839-001`: envtest-backed proof that the production dispatch path (real K8s API + `HandleCreateRR` + `Triager`) propagates `ErrSeverityUndetermined` and creates no RemediationRequest when nothing grounds the resource (closes the CHECKPOINT W gap for this new behavior)
- [x] Confirmed `severity_triage.failed` (AU-2, per `docs/services/apifrontend/security/AUDIT_EVENT_CATALOG.md`) audit-emission coverage is unaffected -- pre-existing `UT-AF-1156-056` proves the audit event fires for any `Triage()` error, including the new terminal `ErrSeverityUndetermined`
- [x] `test/integration/apifrontend`, `test/e2e/apifrontend` -- updated (Tier 2.5/Tier 3-reaching scenarios rewritten to prove new behavior); compiles clean (`go vet`) but **not run locally** -- no envtest/Docker infra available in this sandbox, needs CI verification
- Rewrote all tests that previously exercised Tier 3 (`UT-AF-T-031/032/034/038/043/048/051/051b/051d`, IT `IT-AF-SEV-W01`/`W01b`, E2E `TC-E2E-SEV-04`/`05`) to either prove the new fail-closed behavior or route through Tier 2.5 instead (same underlying `classify()`/`TriageWithRules` LLM call path, just via a correlated rule instead of zero evidence)


jordigilh:
## UT coverage extended to 98.7% (from 86.0%)

Per the coverage classification in `scripts/coverage/coverage_report.py`, `pkg/apifrontend/severity` is the only apifrontend package held to the unit-testable 100% line bar (`pkg/apifrontend/tools` is classified integration-testable, which is why its IT coverage — already extended in the previous commit — is the right lever there, not more UT).

Added targeted UT for previously-uncovered branches across `triage.go`, `anthropic.go`, `llm.go`, `openai_compatible.go`, `cache.go`, and `pod_resolver.go`:
- Nil/empty LLM SDK responses (Anthropic + GenAI)
- Constructor default-value branches (model, logger, bare-`Client`-wrapping paths)
- Cluster-scoped **pending**-alert correlation (`clusterSource`'s second branch)
- Mismatched-value label overlap (proves a shared key with a different value is correctly rejected as a resource-level match)
- Tier 1.5 / Tier 2 rules with unparseable PromQL or non-matching labels
- Tier 2 instant-query failures and the `MaxRulesEvaluated` cap (`break` path)
- `RulesCache` reuse across calls (cache-hit branch)
- `K8sPodResolver` error paths: non-NotFound `Get` errors, missing `spec`/`selector`, malformed selector shape, invalid `matchExpressions` operator, pod-`List` failures
- Context canceled before the pipeline starts

**Package coverage: 86.0% → 98.7%.** The remaining 1.3% is 3 lines that are the literal external-SDK call sites the `ContentGenerator`/`Messager` test seams exist to substitute (`sdkMessager.Create`, `genaiModels.GenerateContent`) plus the GCP-ADC-auth line in `NewAnthropicVertexClient`'s panic-recovery path — exercising these would require live network calls to Anthropic/Vertex AI, which this codebase already carves out via the build-tag-gated `llm_integration_test.go` rather than the default unit run (confirmed empirically that `NewAnthropicVertexClient`'s non-auth branches, e.g. the location-default path, *are* now covered since ADC token resolution is lazy and doesn't block at construction).

All 114 specs pass, `-race` clean across the full `pkg/apifrontend/...` suite, `golangci-lint` clean, `gofmt` clean.

Issue #1941, opened 2026-08-05
Backport #1919/#1940 to release/v1.5: coarse-grained console-access authorization gate


jordigilh:
## Summary

Tracking issue for backporting #1919's fix (a coarse-grained, session-level `kubernaut.ai/console` `use` authorization gate, enforced server-side in every tool-invocation path -- not just an advisory pre-render endpoint) from `main` (#1940) to `release/v1.5`, reverse direction of the usual v1.5-first sequencing (#1912→#1913, #1915→#1917, #1918→#1924, similar direction to #1928), since this was designed and implemented against `main` first.

## Confirmed: the gap exists on `release/v1.5` too

`release/v1.5`'s apifrontend RBAC code predates the `main` refactor and carries the exact same pre-#1919 shape #1919 originally found on `main`:

- `pkg/apifrontend/auth/sar.go`: `SARChecker` only exposes `Check` (`kubernaut.ai/tools`, per-tool). No `ConsoleAuthorizer`/`CheckConsoleAccess`, no `kubernaut.ai/console` resource.
- `pkg/apifrontend/agent/root.go`: `newRBACGuard` (A2A tool-calling path) calls `authorizer.Check(...)` directly with no coarser gate ahead of it -- byte-for-byte the same shape `main`'s `newRBACGuard` had before the #1919 refactor.
- `pkg/apifrontend/handler/mcp_bridge.go`: `checkRBAC` (`/mcp` path) is the same -- calls `cfg.Authorizer.Check(...)` with no console-access precheck.
- `pkg/apifrontend/handler/router.go`: no `/a2a/access` route, no `ConsoleAccessHandler`.
- `charts/kubernaut/values.yaml`/`values.schema.json`: `apifrontend.config.rbac` only has `sarCacheTTL`/`personas` -- no `consoleAccessGroups`.
- `charts/kubernaut/templates/apifrontend/apifrontend.yaml`: only renders `kubernaut-tool-<persona>` ClusterRoles/Bindings -- no `kubernaut-console-access` ClusterRole.

All 6 personas (`sre`, `ai-orchestrator`, `cicd`, `observability`, `l3-audit`, `remediation-approver`) already exist identically in `release/v1.5`'s `values.yaml`, so the persona-name surface the port needs to target is unchanged from `main`.

[...]

Issue #1934, opened 2026-08-04
fleet/mcpclient: ResilientClient.connectWithBackoff has no per-attempt timeout, so a hung SDK call blocks the whole backoff loop forever


jordigilh:
## Context

While triaging #1904 (CI failure from bumping `github.com/modelcontextprotocol/go-sdk` v1.6.1 -> v1.7.0), root-caused in #1933: a wire-protocol handshake change caused `mcp.Client.Connect()` to hang indefinitely against a non-conformant fake server. #1933 fixes the immediate test breakage (fixture-only fix), but it exposed a real robustness gap in production code that is independent of this specific SDK bug.

## The gap

`ResilientClient.connectWithBackoff` (`pkg/fleet/mcpclient/resilience.go`) drives its retry loop with `k8s.io/apimachinery/pkg/util/wait.ExponentialBackoff`, but each attempt calls:

```go
c, connErr := New(ctx, rc.endpoint, rc.opts...)
```

using the **caller-supplied `ctx` directly**, with no per-attempt deadline. `New()` calls `mcpClient.Connect(ctx, transport, nil)`, which can block indefinitely if the underlying SDK/transport never resolves a pending call (as it did here) -- and since production callers of `NewResilient`/`Reconnect` typically pass `context.Background()` (no deadline), a single hung `Connect()` call blocks the entire backoff loop forever, silently defeating `MaxElapsedTime`, `MaxInterval`, and every subsequent retry.

The same unbounded-context issue applies to `ResilientClient.doReconnect`, which also calls `New(ctx, rc.endpoint, rc.opts...)` directly with the caller's context.

## Suggested fix

Wrap each connection attempt in `connectWithBackoff` (and `doReconnect`) with a bounded per-attempt `context.WithTimeout`, derived from the caller's context, so a hung SDK/transport call fails fast and lets the backoff loop retry (or give up per `MaxElapsedTime`) instead of hanging forever. `WithTokenRefreshTimeout`/`WithHTTPClient` already exist as options but only bound the HTTP client's own timeout, not the `Connect()` call as a whole (protocol-level hangs -- like the one in #1933 -- aren't necessarily surfaced as an HTTP client timeout).

## References

- #1904 (dependabot go-sdk bump that surfaced this)
- #1933 (test-fixture fix for the immediate CI breakage; this issue tracks the production robustness follow-up, deliberately scoped out of that PR)
[...]

Issue #1919, opened 2026-08-04
Gate console UI rendering on a dedicated coarse-grained authorization role, not just per-tool SAR


jordigilh:
## Summary

Any user who authenticates to the console (valid OIDC credentials via `oauth2-proxy`) can render the full Kubernaut chat UI today, regardless of whether they're authorized to use *any* Kubernaut tool. Authorization only happens per-tool-call, via `SubjectAccessReview` (SAR) against `kubernaut.ai/tools` (`pkg/apifrontend/auth/sar.go`'s `SARChecker`, invoked from `mcp_bridge.go`/`agent/root.go`) — this check runs *after* the UI has already rendered and the user starts interacting.

Found during review of [#1909](https://github.com/jordigilh/kubernaut/pull/1909) (originally scoped there as audience-bound `TokenReview` for AF/KA, out of scope for that PR — see discussion there). Out of scope for #1909; tracked here for v1.6.

## Problem

- `kubernaut-console`'s `App.tsx` unconditionally renders `<KubernautChat>` once OIDC auth succeeds — there is no authorization gate before the UI mounts.
- Per [ADR-022](docs/services/apifrontend/adr/ADR-022-af-sa-unified-security-model.md), MCP RBAC (SAR on `kubernaut.ai/tools`) is the sole application-level authorization gate, and it's deliberately per-tool/fine-grained — there's no coarser "may this user use the console at all" check.
- Result: a user with a valid credential but zero tool grants gets a fully-rendered chat UI, then every single action they attempt is denied — poor UX (looks broken rather than "you're not authorized"), and reveals the existence/names of internal tools to a user with no legitimate access to any of them.

## Proposed design

Add a coarse-grained, session-level authorization gate that runs *before* the UI renders, layered on top of (not replacing) the existing per-tool SAR checks:

1. **AF**: expose a lightweight endpoint (e.g. `GET /api/v1/console/access`) that performs a single `SubjectAccessReview` against a synthetic resource, e.g. `kubernaut.ai/console` verb `use` — mirroring the existing `kubernaut.ai/tools` verb `use` convention in `deploy/apifrontend/overlays/e2e/e2e-user-rbac.yaml`. Returns `200`/`403` (no per-tool detail needed at this stage).
[...]

Issue #290, opened 2026-03-06
Enhancement: Historical parent-child audit event backfill


jordigilh:
## Summary

DataStorage needs a migration to backfill `parent_event_date` for existing audit events to enable historical event chain queries.

**BR Reference**: BR-STORAGE-037

## Proposed Change

- Create migration to populate `parent_event_date` from existing `parent_event_id` relationships
- Ensure backward compatibility during backfill

## Priority

P2 (Medium) — Planned for V1.1, depends on BR-STORAGE-036 (parent event lookup index).

PR #1942, opened 2026-08-05
fix(apifrontend): backport console-access gate to release/v1.5 (#1919, #1941)

jordigilh:
## Summary

Backports the console-access authorization gate from `main` (issue #1919, PR #1940) to `release/v1.5`, tracked by #1941.

- Adds `SARChecker.CheckConsoleAccess` / `ConsoleAuthorizer`, checking a new coarse-grained `kubernaut.ai/console` `use` SAR grant, separate from the existing per-tool `kubernaut.ai/tools` check (AC-6: least privilege — "console access" and "tool access" are two explicit, independently-auditable grants).
- **The actual security fix**: enforces this gate server-side at both existing tool-invocation paths — `checkRBAC` (`/mcp`) and `newRBACGuard` (`/a2a/invoke`'s tool-calling loop) — so a client that never calls the advisory endpoint below cannot bypass it (AC-3).
- Adds `GET /a2a/access`, a lightweight authenticated pre-flight endpoint the console/chat client can call once to decide whether to render its UI, backed by the same grant. Denials emit `audit.EventAuthAccessDenied` (AU-12).
- Adds the `kubernaut-console-access` Helm RBAC manifests (ClusterRole + one ClusterRoleBinding per group in the new `consoleAccessGroups` values key, defaulted to all six personas) and CI-gated helm-unittest coverage.
- Ports the E2E RBAC overlay with a dedicated, cluster-scoped `e2e-console-access` ClusterRole/Bindings (kept separate from the namespace-scoped `e2e-user-k8s-tools` binding used to prove K8s namespace restriction, per the RCA on the original PR #1940).
- Fixes a pre-existing chart bug found along the way: `app.kubernetes.io/part-of: kubernaut` was set both explicitly and via the `kubernaut.labels` helper on the `kubernaut-tool-*` ClusterRoles, producing a duplicate YAML key that only `helm-unittest`'s strict decode (not `helm template`) caught.

`main` and `release/v1.5` had diverged significantly in the affected files, so this is a surgical re-application of the #1919 diff onto `release/v1.5`'s actual code shape, not a cherry-pick. See `docs/tests/1919/TEST_PLAN.md` §0 for the full list of adaptations (notably: #1934/mcpclient timeout is out of scope here since `pkg/fleet` doesn't exist on this branch).

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/apifrontend/... ./cmd/apifrontend/... -count=1` (full regression, all green)
- [x] `go test ./pkg/apifrontend/auth/... ./pkg/apifrontend/handler/... ./pkg/apifrontend/agent/... -race -count=1`
- [x] `golangci-lint run ./pkg/apifrontend/... ./cmd/apifrontend/...` — 0 issues
- [x] `helm lint charts/kubernaut/ --strict` — clean
- [x] `helm unittest charts/kubernaut` — 18/18 passed (incl. 6 new console-access RBAC cases)
- [x] New unit/integration tests: UT-AF-1919-001..009, IT-AF-1919-001..006, IT-HELM-1919-001..003 — all passing, including the bypass-closure proof (`/a2a/access` never called, tool call still denied)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to Kubernaut will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Security
+
+- **Coarse-grained console-access authorization gate (#1919, #1941, AC-3/AC-6/AU-12)** — Adds a new `kubernaut.ai/console` "use" SubjectAccessReview grant, enforced server-side at both existing AF tool-invocation paths (`/mcp` and `/a2a/invoke`), in addition to the pre-existing per-tool `kubernaut.ai/tools` check. Also adds an advisory `GET /a2a/access` pre-flight endpoint for UI clients. **Operator action may be required**: the chart's new `apifrontend.config.rbac.consoleAccessGroups` value defaults to all six built-in personas (`sre`, `ai-orchestrator`, `cicd`, `observability`, `l3-audit`, `remediation-approver`), so deployments using only those default persona group names need no changes. If you configured a custom group under `apifrontend.config.rbac.personas` (or renamed/replaced the defaults), you must add that same group name to `apifrontend.config.rbac.consoleAccessGroups`, or users in that group will have **every** AF tool call denied after upgrading, even though their existing per-tool grants are unchanged. `helm install`/`helm upgrade` now prints an `NOTES.txt` warning listing any `personas` group missing from `consoleAccessGroups` to catch this case at deploy time.
+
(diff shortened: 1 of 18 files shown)


jordigilh:
Heads-up filed against the operator: this chart's default `consoleAccessGroups` (built-in role-name strings) won't match real operator-managed clusters that customize `spec.apifrontend.rbac.roleBindings` group mapping — confirmed live on our shared dev cluster (all 6 roles collapse to a single custom group, `platform-engineering`, not the literal `sre`/`ai-orchestrator`/etc. defaults). Operator has no CRD support for the new console-access binding at all yet, so this would be a total console outage on upgrade for such clusters. Filed jordigilh/kubernaut-operator#289 (v1.5) and #290 (v1.6 clone).

Issue #1955, opened 2026-08-06
AF: Vertex AI/Anthropic (Claude) LLM streaming call has no client-side timeout — agent loop can hang indefinitely on a stalled model response


jordigilh:
## Summary

While investigating #1949/#1954, confirmed a second, independent missing-timeout gap in AF: **AF's own agent-loop LLM call to Claude (whether via direct Anthropic API or Vertex AI) has no bounded timeout anywhere in the stack**, for the streaming path the interactive agent loop actually uses. This is the same failure class as #1949 (unbounded LLM call, no self-healing) but on AF's model-call boundary rather than KA's, and with *no* backstop at all (unlike #1949, which at least has KA's own inactivity timeout as a side-channel safety net).

Not yet confirmed as the cause of any specific observed hang (the #1949 repro's stuck goroutine pointed at the MCP transport layer instead, tracked in #1954) — filing this because the code-level gap is real, evidenced, and independent of whichever theory explains any one specific incident.

## Evidence (code-level, all three layers unbounded for Claude)

1. **AF's own config knob is silently unwired for this provider.** `LLMConfig.TimeoutSeconds` (`pkg/apifrontend/config/config.go`, defaults to `DefaultLLMTimeoutSeconds = 120`) is only ever applied via `buildLLMHTTPClient`/`BuildLLMHTTPClient`, which is only called from `newGeminiModel` and `newOpenAICompatibleModel` (and, on `main` only, `newVertexGeminiModel`). `newVertexAnthropicModel` / `newAnthropicModel` (`pkg/apifrontend/launcher/model.go`) never call it — confirmed by the code's own comment: *"Vertex AI and Anthropic providers cannot receive a custom transport yet because the ADK wrapper (adk-anthropic-go) does not expose HTTP client injection"* (referencing #1342, still open on both `release/v1.5` and `main`).
2. **The ADK wrapper's streaming path has no timeout logic of its own.** `adk-anthropic-go`'s `generateStream` calls `anthropic-sdk-go`'s `Messages.NewStreaming`, which — unlike the non-streaming `Messages.New` (which self-bounds via `CalculateNonStreamingTimeout`, up to 1 hour by default) — sets no `option.WithRequestTimeout` at all.
3. **The underlying SDK's default HTTP client has no timeout.** `anthropic-sdk-go`'s `default_http_client.go`: `return &http.Client{Transport: http.DefaultTransport}` — `Timeout` field is left at its zero value (unset).
[...]

Issue #1954, opened 2026-08-06
AF: pooled MCP tool calls have no client-side timeout — lost/stuck response on a reused session blocks discover_workflows for up to 10 minutes with no error surfaced


jordigilh:
## Summary

Follow-up to #1949. The `#1953` diagnostic re-repro (see [this comment](https://github.com/jordigilh/kubernaut/issues/1949#issuecomment-5199395443)) proved KA's side of the #1949 fix (#1951) is working correctly — KA computes and returns a valid `discover_workflows` response every time, and its own inactivity timeout correctly self-heals a stuck session. The remaining hang is entirely on AF's side, and this issue narrows it to a specific, previously-known-risky area: **`KASessionPool` reused ("cache hit") connections have zero client-side timeout on the tool-call round trip**, so a response that's lost or delayed on the wire blocks AF's MCP client forever (or until KA's own ~10-minute inactivity timeout tears down the transport out from under it).

## Evidence (from #1949's latest repro)

For every hung RR in that run, AF's log for the affected `rr_id` is:

```
"tool call started"              tool=kubernaut_discover_workflows
"pool session reused (cache hit)"
```

...and then **nothing else, ever**, for that `rr_id` — no timeout, no error, no retry. A `SIGQUIT` goroutine dump captured on a still-live occurrence of this failure mode shows AF's MCP client genuinely blocked reading the next message:

```
goroutine 33152 [select]:
github.com/modelcontextprotocol/go-sdk/mcp.(*streamableClientConn).Read(...)
	.../mcp/streamable.go:2205
github.com/modelcontextprotocol/go-sdk/internal/jsonrpc2.(*Connection).readIncoming(...)
	.../internal/jsonrpc2/conn.go:517
```

Meanwhile KA's side (per #1953's step-level tracing) shows the handler completing and returning the marshaled response in well under 100ms after the LLM turn — including at least one occurrence where a **real, valid workflow was computed and then silently discarded**, with KA's own completion bookkeeping later misreporting `has_workflow:false` once the inactivity timeout fired.

## Root cause chain (why there's no timeout here)

1. #1386 (closed) fixed a *different* bug — pooled sessions going stale because the shared `http.Client.Timeout` (30s) was killing idle SSE streams — by giving long-lived MCP connections a client with **no `Timeout` at all**, relying on the SDK/context to manage stream lifecycle instead.
[...]

Issue #1949, opened 2026-08-05
workflow_discovery phase can hang indefinitely with no timeout after LLM's terminal decision response (reproduced 2x/run)


jordigilh:
## Summary

An interactive investigation's `workflow_discovery` phase can hang **indefinitely with no timeout, no error, and no audit trail** after the LLM produces its terminal decision-bearing response. The session is only ever cleaned up by KA's wall-clock inactivity timer (~10 min later), which is fully decoupled from the actual stuck goroutine — the CR transitions to `Failed`/`TimedOut` but the underlying hang is never resolved (the goroutine leaks). Reproduced **twice independently in the same ~30-minute test run**, on two different targets (one `crashloop`-based, one `memory-eater`-based), so this is not target-specific flakiness.

Deployed image: `ghcr.io/jordigilh/kubernaut/kubernautagent:main-cc319eed729058710a695384b5f1ebc795a435a1` — confirmed to be `release/v1.5` HEAD, which **includes** PR #1947/#1945's workflow-discovery message-propagation fix. This is not that issue resurfacing; it's a distinct, new-to-this-run defect (see RCA below for why #1947's diff was ruled out).

## Reproduction (kubernaut-console live E2E suite, `playwright.live-v15.config.ts`, real Sonnet 5, no mock-llm)

Two RRs from a single suite run on 2026-08-05, both interactive/console-driven sessions:

| RR | Target | Session ID | Last activity | Outcome |
|---|---|---|---|---|
| `rr-868e1536a0a2-a87819be` | `console-e2e-lifecycle-3/Deployment/worker` (crashloop scenario) | `d9c6b378-eb68-4f0a-8661-93a0d65bc4c3` | 2026-08-05T17:55:26Z | `Failed`, `has_workflow:false`, teardown at 18:05:35Z (exactly 10m10s of silence) — this one **caused a visible test failure** (Playwright timed out waiting for a workflow card) |
| `rr-777f63b4f29b-c8eceeb8` | `console-e2e-full-remediation-3/Deployment/memory-eater` | `02283903-7aa0-490b-a529-9d1e9ee346aa` | 2026-08-05T17:52:44Z | `TimedOut`, `has_workflow:false`, teardown at 18:02:44Z (exactly 10m00s of silence) — the *test* itself had already passed on an earlier assertion before this session hung in the background |

### Audit trail pattern (identical in both, from `audit_events` via `kubernaut-demo-scenarios/scripts/capture-eval.sh`)

[...]

Issue #1342, opened 2026-06-01
LLM transport: add mTLS client certificates and standardize transport chain across KA and AF


jordigilh:
## Summary

The LLM HTTP transport chain has gaps that block enterprise deployment scenarios where a proxy or gateway sits between kubernaut and the LLM provider. This issue proposes three related changes to standardize the transport stack across Kubernaut Agent (KA) and API Frontend (AF).

### 1. No mTLS client certificate support

`tlsCaFile` configures server certificate verification (one-way TLS), but there is no way to present a client certificate. In airgapped/enterprise environments, an mTLS proxy often sits between the pod and the LLM provider, requiring the client to authenticate with a certificate.

**Affected:** Both KA and AF — `NewTLSTransport()` in `pkg/shared/tls/tls.go` only sets `RootCAs`, not `Certificates`.

### 2. AF rejects transport options for `vertex_ai` and `anthropic`

AF validation in `pkg/apifrontend/config/config.go` explicitly blocks `tlsCaFile`, `oauth2`, `customHeaders`, and `circuitBreaker` for `vertex_ai` and `anthropic` providers. This means AF cannot:

- Authenticate to an OAuth2 corporate LLM gateway when using these providers
- Use a custom CA or mTLS when connecting through a proxy
- Inject custom headers (e.g. tenant ID, tracing)

KA has the same limitation for `vertex_ai` (the vertexanthropic client constructs its own HTTP client internally, bypassing the transport chain).

**Current AF transport support matrix:**

| Capability | `vertex_ai` | `gemini` | `anthropic` |
|---|---|---|---|
| `tlsCaFile` | Rejected | Yes | Rejected |
| OAuth2 | Rejected | Yes | Rejected |
| Custom headers | Rejected | Yes | Rejected |
| Circuit breaker | Rejected | Yes | Rejected |

### 3. Transport chain is not uniform across components

KA and AF should apply the same transport chain order for all providers:

```
CircuitBreaker → CustomHeaders → OAuth2 → mTLS/TLS → base HTTP
```

This chain wraps the HTTP client that the provider SDK uses, so it is orthogonal to the provider implementation (LangChainGo, Anthropic SDK, ADK, etc.).

## Deployment scenarios blocked

1. **mTLS proxy (no kagenti):** A TLS-terminating proxy requires client certificates. Neither KA nor AF can present them.
[...]

PR #1960, opened 2026-08-06
fix(af): bound pooled MCP tool calls and Claude/Vertex-Claude LLM calls with client-side timeouts

jordigilh:
## Summary

Closes two independent, already-root-caused hangs on `release/v1.5`, both surfaced during the #1949 re-repro investigation and confirmed by a full client-side-timeout audit of AF/KA's A2A and MCP call paths.

- **#1954** — none of AF's ADK-facing interactive tool handlers (`discover_workflows`, `select_workflow`, `message`, `complete`, `cancel`, `status`, `reconnect`) bound their call to the pooled KA MCP client with any deadline. A lost/stuck response on a reused session blocked the caller indefinitely. `mcp_bridge.go`'s existing 30s tool timeout protects an unrelated, separately-exposed MCP-server surface, not this path (see [correcting comment on #1954](https://github.com/jordigilh/kubernaut/issues/1954#issuecomment-5204028540)).

  Fix: add `PooledToolCallTimeout` (60s, overridable var) and wrap the three call sites — `HandleDiscoverWorkflows`, `HandleSelectWorkflow` (`ka_tools.go`), and `invokeInteractiveAction` (`ka_interactive.go`, shared by the 5 remaining tools) — with `context.WithTimeout(ctx, PooledToolCallTimeout)`.

- **#1955** — AF's own agent-loop LLM call to Claude (direct Anthropic API or Vertex AI) had no bound anywhere. `adk-anthropic-go` doesn't expose HTTP client injection (#1342), so the fix is a context-deadline decorator around `model.LLM` rather than an HTTP client change.

  Fix: add a `timeoutModel` decorator that wraps every `GenerateContent` call with `context.WithTimeout(ctx, m.timeout)`, resolving the timeout from `cfg.TimeoutSeconds` (falling back to `DefaultLLMTimeoutSeconds`) — the same resolution `buildLLMHTTPClient` already uses for the Gemini/OpenAI-compatible paths, just applied at the context layer. Applied in `newAnthropicModel` and `newVertexAIModel`.

## Pyramid invariant

Both fixes are proven end-to-end, not just in isolation — each gets a UT for the decorator's own logic plus an IT that drives the real production call chain with only the genuine external boundary faked/soft-checked:

| Component | Production entry point | IT |
|---|---|---|
| Pooled MCP call timeout | `HandleDiscoverWorkflows`/`HandleSelectWorkflow`/`invokeInteractiveAction` | `IT-AF-1954-004` — real `ka.KASessionPool` + `ka.PooledMCPClient`, only the network-facing `PoolSession` faked |
| Claude/Vertex-Claude LLM call timeout | `launcher.NewModelFromConfig` | `IT-AF-1955-003`/`IT-AF-1955-004` — real construction, `IsTimeoutWrappedForTest` asserts the decorator is actually applied |

No new production entry points are introduced; the value of both fixes is that a deadline now propagates through and is enforced by an *existing* real call chain.

## Test plan

- [x] `golangci-lint run` on changed packages — 0 issues
- [x] New specs (`UT-AF-1954-001..003`, `IT-AF-1954-004`, `UT-AF-1955-001..002`, `IT-AF-1955-003..004`) pass
- [x] Full `pkg/apifrontend/...` suite passes (23/23 packages, zero regressions)

Refs #1954, #1955

### DIFF
--- a/pkg/apifrontend/launcher/export_test.go
+++ b/pkg/apifrontend/launcher/export_test.go
@@ -3,9 +3,11 @@ package launcher
 import (
 	"context"
 	"net/http"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/a2aproject/a2a-go/a2a"
+	"google.golang.org/adk/model"
 	"google.golang.org/adk/server/adka2a"
 	"google.golang.org/adk/session"
 	"google.golang.org/genai"
@@ -105,4 +107,17 @@ func EmitArtifactForTest(ctx context.Context, data map[string]any, textFallback 
 // ValidatePayloadForTest exports ValidatePayload for testing.
 func ValidatePayloadForTest(schemaName string, data map[string]any) error {
 	return ValidatePayload(schemaName, data)
+}
+
+// IsTimeoutWrappedForTest reports whether m is a *timeoutModel (the #1955
+// decorator applied by newAnthropicModel/newVertexAIModel) and, if so, its
+// configured timeout. Lets black-box (launcher_test) specs prove the
+// decorator is wired into NewModelFromConfig's real construction sites
+// without exposing timeoutModel itself.
+func IsTimeoutWrappedForTest(m model.LLM) (time.Duration, bool) {
+	tm, ok := m.(*timeoutModel)
+	if !ok {
+		return 0, false
+	}
+	return tm.timeout, true
 }
--- a/pkg/apifrontend/launcher/model.go
+++ b/pkg/apifrontend/launcher/model.go
@@ -19,6 +19,7 @@ package launcher
 import (
 	"context"
 	"fmt"
+	"iter"
 	"net/http"
 	"os"
 	"strings"
@@ -93,7 +94,11 @@ func newVertexAIModel(ctx context.Context, cfg config.LLMConfig) (m model.LLM, e
 	if cfg.Endpoint != "" {
 		adkCfg.BaseURL = cfg.Endpoint
 	}
-	return adkanthropic.NewModel(ctx, anthropic.Model(cfg.Model), adkCfg)
+	llm, err := adkanthropic.NewModel(ctx, anthropic.Model(cfg.Model), adkCfg)
+	if err != nil {
+		return nil, err
+	}
+	return wrapWithTimeout(llm, cfg), nil
 }
 
 func newGeminiModel(ctx context.Context, cfg config.LLMConfig) (model.LLM, error) {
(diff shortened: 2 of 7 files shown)


jordigilh:
## Validated on shared OpenShift dev cluster — fix resolves the hang

Deployed `apifrontend:pr-1960` (this PR's image, `linux/amd64,linux/arm64`) to our shared dev cluster, keeping `kubernautagent:pr-1953` (the #1949 diagnostic step-tracing build) in place so KA's fine-grained `discover_workflows:` logs would still corroborate the result.

**Method**: re-ran the same 4-concurrent-interactive-session repro that previously reproduced the hang (2/4 sessions stuck indefinitely). First pass reused stale fixture namespaces from earlier rounds and all 4 "completed" in ~13s — traced that to AF's `session_active` cache short-circuit (not exercising this code path at all), so discarded it and re-ran against **brand-new, never-touched fixture namespaces** to force genuinely fresh interactive investigations.

**Result — 4/4 completed, no hang:**
```
[1] COMPLETED for console-e2e-approval-2-203db3
[0] COMPLETED for console-e2e-approval-203db3
[3] COMPLETED for console-e2e-lifecycle-3-203db3
[2] COMPLETED for console-e2e-contract-3-203db3
```

**KA-side log evidence** (from the #1953 diagnostic build, same window) — every `discover_workflows` call now has a matched start → completion pair, with no orphans:

```
12:47:14.324  discover_workflows: using stored RCA result from autonomous investigation
12:47:40.223  discover_workflows: returning response to caller        (~26s)

12:49:17.533  discover_workflows: using stored RCA result from autonomous investigation
12:49:30.659  discover_workflows: returning response to caller        (~13s)

12:49:40.261  discover_workflows: using stored RCA result from autonomous investigation
12:49:56.352  discover_workflows: returning response to caller        (~16s)

12:49:51.385  discover_workflows: using stored RCA result from autonomous investigation
12:50:10.052  discover_workflows: returning response to caller        (~19s)
```

All 4 concurrent calls completed well inside the new `PooledToolCallTimeout` (60s), and AF's own logs confirm the caller actually received each result (`kubernaut_present_decision` tool calls fire immediately after each `discover_workflows` return, meaning the LLM agent loop received and acted on the response — the exact thing that never happened in the hung reproductions).

This is the "different location, same #1949 hang" behavior from the round-3 re-repro — this PR's pooled-MCP-call timeout closes it. From our validation, this is ready to merge.

Issue #1998, opened 2026-08-07
apifrontend: phaseGuardAfter short-circuits AfterToolCallback chain, silently dropping tool-call-completion logging (discovered via #1995)


jordigilh:
## Summary

`phaseGuardAfter` (`pkg/apifrontend/agent/phase_guard.go`) violates ADK's documented `AfterToolCallback` chaining contract, silently suppressing every callback registered after it in the chain — currently just `afterLog` (tool-call-completion/timeout observability logging), but it would silently swallow any future callback too.

Discovered while investigating #1995 (this is a separate, independent defect — not the cause of that hang, but it did block us from confirming whether a timeout actually fired during that investigation).

## Root cause

ADK's `invokeAfterToolCallbacks` (`google.golang.org/adk`, `internal/llminternal/base_flow.go`) is documented and implemented as a "first non-nil result wins" chain:

```go
for _, callback := range f.AfterToolCallbacks {
    result, err := callback(toolCtx, tool, fArgs, fResult, fErr)
    if err != nil {
        return nil, err
    }
    // When a list of callbacks is provided, the callbacks will be called in
    // order they are listed while a callback returns nil.
    if result != nil {
        return result, nil
    }
}
```

A callback must return `(nil, nil)` to "pass through" and let the chain continue; returning non-nil is an explicit signal to override the result and stop.

`root.go` registers `AfterToolCallbacks: []llmagent.AfterToolCallback{afterMetrics, afterAudit, afterPhase, afterLog}`. `afterMetrics` and `afterAudit` correctly return `(nil, nil)` unconditionally. `afterPhase` (`phaseGuardAfter`) does not:

```go
// pkg/apifrontend/agent/phase_guard.go
if isDiscoverWorkflows {
    recordDiscoverWorkflowsCheckpoint(ctx)
    return resp, callErr   // resp is the original, non-nil fResult
}
...
if isEntry { recordDriverEntryState(...) }
if isTerminal { clearDriverSessionState(ctx) }
syncActiveContextRegistry(...)
return resp, callErr       // same problem
```

Since `resp` (== `fResult`) is non-nil for any tool call that produced a structured result, this unconditionally short-circuits the chain and **`afterLog` never runs** for any successful call to `kubernaut_discover_workflows`, `kubernaut_investigate`, `kubernaut_reconnect`, `kubernaut_complete`, `kubernaut_cancel`, or `kubernaut_complete_no_action`.

## Impact

[...]

Issue #1995, opened 2026-08-07
PooledToolCallTimeout (#1960) doesn't unblock discover_workflows when KA's MCP response is lost — hang still bounded only by KA's 10-min inactivity timeout


jordigilh:
## Summary

`workflow_discovery`'s `no_matching_workflows` path can still hang for the full ~10 minutes despite #1960's `PooledToolCallTimeout` (60s), because the AF-side context deadline never actually unblocks the pooled MCP `CallTool()` invocation when KA's response is lost in transit. The system falls all the way back to KA's own `interactive.inactivityTimeout` (~10 min) as the *only* effective backstop — the exact scenario #1954/#1960 were meant to protect against.

This reproduced **4/4 times** across two independent test runs (2 pre-existing, 2 freshly reproduced with KA pprof profiling enabled for this report) against `release/v1.5` @ **v1.5.6-rc2** (confirmed via image digest, which includes #1960's merge commit `a5b7d15c`).

## Why this isn't just #1949/#1950/#1954 reopened as-is

Those issues are closed with a validated fix (#1960, `PooledToolCallTimeout = 60s` wrapping the pooled-session MCP call sites in `pkg/apifrontend/tools/ka_tools.go` / `ka_interactive.go`). That validation showed 4/4 concurrent sessions completing in 13–26s. This report shows that when the failure mode #1949 originally described actually occurs (KA's MCP response silently lost on a reused/pooled session), **the 60s context deadline does not fire** — so the fix provides no observable improvement for the case it was built for. I'm filing this separately (rather than reopening #1949) since the RCA is more precise now and the closure criteria for #1949 (clean 4/4 repro) technically were met for the *load-trigger* scenario validated there — this is the narrower "deadline doesn't actually cancel the blocked read" gap.

## Environment

- `kubernaut` / `apifrontend` / `kubernautagent`: **v1.5.6-rc2**, confirmed by digest match against `quay.io/kubernaut-ai/apifrontend:1.5.6-rc2` and `quay.io/kubernaut-ai/kubernautagent:1.5.6-rc2`.
- `kubernaut-operator`: v1.5.8-rc1.
- Model: `claude-sonnet-4-6`.
- Found via `kubernaut-console`'s live E2E suite (`e2e/live/approval-gate.spec.ts`, `playwright.live-v15.config.ts`) against a shared OpenShift dev cluster, real (non-mock) LLM calls.
[...]

Issue #1899, opened 2026-08-03
AF system prompt: 'Alert Prioritization' section overrides consent gate, launches unsolicited investigations on read-only 'list alerts' queries


jordigilh:
## Summary

The AF agent's system prompt (`pkg/apifrontend/agent/prompt.txt`) contains a direct, self-contradicting instruction that causes the LLM to autonomously launch a real investigation — consuming real LLM cost and creating a decision-pending artifact requiring user action — in response to a purely read-only query like "list active alerts", with no explicit user consent.

## Root cause

Three separate places in the prompt establish a consent gate for observation-triggering phrasing ("list", "show", "status"):

1. `### Observation Mode (read-only)` (prompt.txt line ~52): *"Do not initiate investigations or remediations... do NOT resume, investigate, or act on them unless the user EXPLICITLY requests it."*
2. `roleGuidanceMap["sre"]` (`pkg/apifrontend/agent/prompt.go`): *"Do NOT proactively investigate or act on existing incidents unless the user explicitly requests it."*
3. The `Decision Algorithm` section itself: *"Is the user asking a simple informational question (list, status, show)? → Observation tools ONLY. Do NOT investigate or remediate."*

But the `## Alert Prioritization` section (prompt.txt line ~198-207) unconditionally overrides all three:

> When `list_alerts` returns a `prioritized` field... `selected_index`: Index of the highest-severity, longest-firing alert. **Investigate this one.**
> ...
> **Always use `kubernaut_investigate_alert` with `alerts[selected_index]` unless the user explicitly picks a different one from `tied_indices`.**

This directive has no gate on user intent/mode — it fires purely off the shape of the `list_alerts` tool response, regardless of whether the user asked to investigate or merely to list. Confirmed present identically on both `main` (`a8b9edd4f`) and `release/v1.5`.

## Reproduction (observed live, 2026-08-03, shared OpenShift dev cluster)

1. User opens the console chat for an existing RR (`rr-44a43cbae6ed-20c310cf`, alert `KubePodCrashLooping`, namespace `console-e2e-lifecycle-2`).
2. User types **"list active alerts"** — a purely informational request, matching the Observation Mode trigger phrase "list" exactly.
3. LLM correctly calls `list_alerts` first (read-only, as instructed).
[...]

Issue #1716, opened 2026-07-23
AF: relay a live signal for redacted reasoning turns (BR-AI-086) — currently a full no-op, Console has nothing to render


jordigilh:
## Background

`kubernaut-console#32` (https://github.com/jordigilh/kubernaut-console/issues/32) tracks Console-side work to render KA's captured LLM reasoning/thinking content (BR-AI-086, #1634/#1635) in `ThinkingPanel`. Part of that scope is a placeholder for the case where a provider redacts its reasoning (e.g. Anthropic `redacted_thinking`) — reasoning occurred, but its content is opaque.

While scoping that Console work, I traced the `Redacted` flag through the full pipeline and found there's currently **no live signal for Console to render a placeholder against**:

- `pkg/kubernautagent/llm/anthropicfamily/client.go` correctly captures `redacted_thinking` blocks (`ReasoningBlock.Redacted = true`, empty `Text`).
- `pkg/apifrontend/tools/ka_investigate_bridge.go:296-301` (`FormatEventForUser`, case `EventTypeReasoningContentDelta`) extracts only the `text` field — on a redacted turn this is `""`.
- `pkg/apifrontend/launcher/event_bridge.go:244-246` (`EmitReasoningContent`) forwards that into `emitWithLimit`, which **no-ops on empty text** (line 253-255), by explicit design per the comment at 241-243 and `reasoning_content_relay_1635_test.go:110-127` ("redacted (empty-text) reasoning content must be a no-op on the AF relay ... the audit trail remains the record of truth for redacted=true").

Net effect: when a turn's reasoning is redacted, **zero SSE bytes are emitted** — not an event carrying `redacted:true`, nothing at all. The only place `Redacted` is preserved today is the audit-trail type (`katypes.ReasoningSummary`), which Console never sees live. From the operator's perspective, a redacted turn is indistinguishable from "no reasoning content produced this turn" — which reads as a silent gap rather than expected provider behavior.

## Proposal

Have AF emit a live, content-free signal on a redacted turn instead of a full no-op — e.g. still call through to the SSE bridge with `metadata.redacted=true` (and no/empty `text`) rather than short-circuiting at `emitWithLimit`'s empty-text guard. This carries zero reasoning content (nothing opaque leaks), just the boolean fact that redaction occurred, so Console can render "Reasoning hidden by provider" instead of nothing.

Open design questions (not resolving here, flagging for the team that owns this pipeline):
[...]

PR #2001, opened 2026-08-07
Fix #1996 + #1998 + #2013 + #2006 + #2012 + #2016 + #2021 + #2020: pprof gate, phaseGuardAfter audit-log, KA session auto-close, reasoning dedup, RO timeout race, KA MCP logger, cluster-alert mis-correlation, KA finalization gap (main port of PR #2000)

### LINKED ISSUE #1996: [v1.6] PooledToolCallTimeout (#1960/#1963) doesn't unblock discover_workflows when KA's MCP response is lost — hang still bounded only by KA's 10-min inactivity timeout
v1.6/main tracking clone of #1995 (same root cause, `main` code path). See #1995 for the full RCA, reproduction evidence, goroutine dumps, and suggested fix direction — duplicating the summary here so both release lines can track/schedule the fix independently, mirroring how #1950 tracked #1949.

## Summary

`workflow_discovery`'s `no_matching_workflows` path can hang for the full ~10 minutes despite `PooledToolCallTimeout` (60s, from #1963 — the `main` port of #1960/#1954's fix), because the AF-side context deadline never actually unblocks the pooled MCP `CallTool()` invocation when KA's response is lost in transit. The system falls all the way back to KA's own `interactive.inactivityTimeout` (~10 min) as the only effective backstop — the exact scenario #1954/#1960/#1963 were meant to protect against.

Not yet independently reproduced against a `main` deployment (this clone is filed on code-inspection confidence, not a live repro — see below); reproduction in #1995 was against `release/v1.5` @ v1.5.6-rc2.

## Why I'm confident this affects `main` too

Confirmed by direct code inspection of `origin/main` @ `235fae0` (2026-08-07):

- #1963 (`main` port of #1960) is merged into `main` (merged 2026-08-07T01:10:57Z, PR base `main`).
- `pkg/apifrontend/tools/ka_tools.go` on `main` has byte-identical `PooledToolCallTimeout` logic to `release/v1.5` — same `context.WithTimeout(ctx, PooledToolCallTimeout)` wrapping at the same line numbers (88, 265), same `context.DeadlineExceeded`/`context.Canceled` handling.
- `pkg/apifrontend/ka/pooled_mcp_client.go`'s `DiscoverWorkflows` → `callPooledTool` → `doCallTool` → `session.CallTool(ctx, ...)` call chain is unchanged.
- `go.mod` on `main` pins the identical `github.com/modelcontextprotocol/go-sdk v1.7.0` as `release/v1.5` — so if the root cause is in `go-sdk`'s `ClientSession.CallTool` not honoring per-call context cancellation on a shared `jsonrpc2.Connection` read loop (the leading hypothesis in #1995), `main` is equally exposed.
- KA-side `internal/kubernautagent/investigator/investigator.go`'s `submit_result_no_workflow` sentinel classification and `internal/kubernautagent/mcp/tools/event_log_bridge.go`'s `EventLogBridge.Run` are structurally unchanged on `main`.

Given the fix for this is likely a single shared root cause (either in kubernaut's own context wiring or in the vendored `go-sdk` version), fixing #1995 on `release/v1.5` should very likely resolve this clone too — but filing separately per the existing #1949/#1950 convention so `v1.6`/`main` scheduling isn't blocked on `v1.5`'s release train, and so a `main`-side repro can be tracked/closed independently once #1995's fix is ported.

## Suggested validation

Once #1995 lands on `release/v1.5`, port to `main` and re-run the same repro (`kubernaut-console`'s `e2e/live/approval-gate.spec.ts` against a `main`-tracking deployment, or the forced-lost-response test suggested in #1995) to confirm the fix also closes this path on `main` before closing this clone.

## Related

- #1995 — primary issue, full RCA, goroutine-dump evidence, reproduction against `release/v1.5` v1.5.6-rc2.
- #1949 (closed, v1.5) / #1950 (closed, v1.6 clone) — original hang discovery and fix, which this issue shows was insufficient for this specific failure mode.
- #1954 / #1960 / #1963 — the AF-side `PooledToolCallTimeout` fix (v1.5 + main ports) whose effectiveness this issue calls into question.

jordigilh:
## Summary

`main` port of PR #2000 (release/v1.5), which now bundles **eight** independent fixes, all surfaced during the same #1995/#1996 (`workflow_discovery` ~10-minute hang) live-diagnostics investigation and QE's dev-cluster validation of PR #2000. Bundled into one PR here too, to minimize review/merge overhead.

### 1. #1996 diagnostic/hardening groundwork (v1.6 clone of #1995)

- **`ServerConfig.DisableProfiling`** (default `true`) gates `/debug/pprof/*` on the internal health mux only, mirroring `kubernautagent`'s existing secure-by-default posture. Never exposed on the external API router.
- **`closeSessionBounded`**: `PoolSession.Close()` now runs in a background goroutine with a 5s bound, preventing one poisoned session from wedging `EvictIdle`/`DrainAll`/`Inject`/`InjectWithCleanup` for every other pool entry forever.
- **Timeout-fired logging**: `HandleDiscoverWorkflows`/`HandleSelectWorkflow`/`HandleCompleteNoAction` now log a warning when `PooledToolCallTimeout` fires.
- **`HandleCompleteNoAction`** is now itself bounded by `PooledToolCallTimeout`.

Ported from release/v1.5's version of these files, which had diverged significantly (main already has #1899/#1918/#1912 refactors decomposing `phaseGuardAfter` into named helper functions -- see part 2 below).

### 2. #1998: phaseGuardAfter short-circuits the AfterToolCallback chain (v1.6 clone of #1997)

Discovered while tracing why "tool call completed"/timeout-fired logs never showed up during the #1995/#1996 dev-cluster diagnostics above.

- `phaseGuardAfter` re-returned the original `(resp, callErr)` from every exit path instead of `(nil, nil)`. ADK's `invokeAfterToolCallbacks` (`google.golang.org/adk`, `internal/llminternal/base_flow.go`) treats any non-nil result from a callback as an explicit override and stops the chain immediately — so `afterLog` (registered after `afterPhase` in `root.go`'s `AfterToolCallbacks: []llmagent.AfterToolCallback{afterMetrics, afterAudit, afterPhase, afterLog}`) never ran for **any** tool call, silently dropping the `"tool call completed"` audit line.
- `phaseGuardAfter` only ever records a side effect (session state, the active-context registry) — it never needs to modify `resp`/`callErr` — so `(nil, nil)` is the correct "pass through unchanged" signal, exactly matching how `afterAudit`/`afterMetrics` already behave in the same chain.
- `main`'s `.golangci.yml` (unlike release/v1.5's) enables the `nilnil` linter, so the fix adds a function-level `nolint:nilnil` justification matching the existing convention already on `phaseGuardBefore` in this same file.
- **BR-AUDIT-005** (audit completeness) / **FedRAMP AU-3** (content of audit records), **AU-12** (audit generation).

### 3. #2013: KA interactive session isn't released on no_matching_workflows (v1.6 clone of #2003)

Discovered by QE against PR #2000's dev-cluster deployment: a live goroutine dump pinpointed AF's `WatchTerminalEvents` stuck in a `select` for 7+ minutes, waiting on a `session_ended` terminal event that KA never sent.

- KA's interactive session lease was never released when `workflow_discovery` concluded with `HumanReviewReason: "no_matching_workflows"` unless a human/agent explicitly followed up with `complete_no_action`. The session sat "active" until KA's 10-minute `interactive.inactivityTimeout` eventually tore it down, and that inactivity teardown was the *only* thing emitting the `session_ended` event AF's `WatchTerminalEvents` (no timer-based safety net, #1438) was waiting on.
- Widened `AutonomousSessionLifecycle` (`investigate.go`) with `EmitSessionEndedByRR(rrID, reason string)`; added a no-op stub to `NopAutonomousManager` and updated the 10 test doubles implementing the interface.
- `handleDiscoverWorkflows` now calls a new `autoCloseOnNoMatchingWorkflows` helper after Step 4 (`RunWorkflowDiscovery`): when the result carries `HumanReviewNeeded && HumanReviewReason=="no_matching_workflows"`, a background goroutine (re-acquiring the session mutex, KA-HIGH-3) calls `autoMgr.EmitSessionEndedByRR` **before** `CompleteHTTPSession`/`sessions.Release` — the same #1438 ordering already used by `cmd/kubernautagent/routes.go`'s inactivity-timeout and disconnect handlers, so `EventLogBridge` can forward `session_ended` to AF before the channel closes.
- Ported from release/v1.5's #2003 fix, adapted to `main`'s ISP-split `AutonomousSessionQuerier`/`AutonomousSessionLifecycle` interfaces and `investigate_discovery.go`'s structure (GO-ANTIPATTERN-AUDIT-2026-07-01 Phase 5 refactor).
- **BR-INTERACTIVE-010**, **AU-3** (release reason on the terminal event), **SI-4** (event ordering/sequencing).

### 4. #2006: duplicate reasoning_content_delta echoing the narration text (main equivalent of #2010)

Reported after QE observed the Console's Thinking panel showing the same LLM reasoning sentence twice, back to back — same symptom as release/v1.5's #2010, but a **different** underlying mechanism on `main` (main streams reasoning via two separate event types, `EventTypeReasoningDelta` + `EventTypeReasoningContentDelta`, rather than v1.5's single concatenating `reasoningDeltaText` function).

- Claude's extended-thinking + tool-use behavior frequently ends the private thinking block with a short one-line plan and then repeats that identical line as the visible `Content` right before the `tool_use` block. KA faithfully streamed both as separate live events with identical text.
- `emitReasoningContentEvent` now takes the narration text already emitted at each of its three call sites (`investigator_loop.go`, `investigator_rca.go`, `investigator_workflow_selection.go`) and skips `EventTypeReasoningContentDelta` when `Reasoning.Text` matches it after trimming whitespace — no information is lost since the text is already visible via the narration event.
- The redacted-reasoning transparency signal (empty `Text` + `Redacted=true`, #1716) is explicitly exempted from suppression so it always surfaces even when narration also happens to be empty.
- **BR-AI-086**, **FedRAMP CC7.2** (audit trails visible to the operator) **/SI-10** (input validation before display).

### 5. #2012: RemediationOrchestrator discards a same-tick Completed/Failed AIAnalysis (main port of release/v1.5's #2011)

Discovered while confirming #2011's fix was `release/v1.5`-only and auditing `main` for the same bug: `checkPhaseTimeouts`/`applyInteractiveTimeoutExtension` were byte-for-byte identical to `release/v1.5`'s pre-fix versions on `main`, just relocated into `internal/controller/remediationorchestrator/timeout_management.go` by an unrelated file-split refactor (#1520).

- `applyInteractiveTimeoutExtension` recomputes the effective `Analyzing`-phase timeout from `AIAnalysis`'s *current* state on every reconcile. Once `AIAnalysis.Status.InteractiveSession.CompletedAt` is set (session finished), the interactive-session extension is revoked — in the *same* reconcile where a terminal (`Completed`/`Failed`) `AIAnalysis` most needs the extended window to be consumed by the `Analyzing` phase handler, racing a valid `AutoApproved` result into a spurious `TimedOut`.
- `applyInteractiveTimeoutExtension` now also returns `aiTerminal bool` (using the pre-existing `isAIAnalysisTerminal` helper from `cascade_terminal.go`, same package) so `checkPhaseTimeouts` can skip the timeout comparison entirely and let the same-reconcile `Analyzing` phase handler consume the result.
- **BR-ORCH-028** (per-phase timeouts) / **DD-INTERACTIVE-002** (dynamic extension), **FedRAMP IR-4/SI-4** — identical framing to #2011, since this is the same bug at the same logic, just a different file location on `main`.

### 6. #2016: KA MCP server never wires a logger (main port of release/v1.5's #2017)

Discovered alongside #2012 while auditing `main` for the remaining #1995-adjacent gaps: `NewMCPServer` never set `ServerOptions.Logger`, so go-sdk-internal log/error events (including `OnInternalError`, relevant to upstream go-sdk#1147/#1153) were silently discarded by the SDK's default discard logger.

- `NewMCPServer` now accepts variadic `MCPServerOption`; the new `WithMCPLogger` option forwards a `logr.Logger` through the `logr`→`slog` adapter (`logr.ToSlogHandler`) into `mcpsdk.ServerOptions.Logger`.
- `MCPDeps.Logger` is threaded through `BootstrapMCP` from `cmd/kubernautagent/routes.go`'s `buildMCPHandler` (the equivalent of `release/v1.5`'s `main.go` call site, relocated by an unrelated decomposition) — the existing production call site now supplies a real logger instead of leaving it unset.
- **AU-2/AU-3** (Audit Events / Content of Audit Records), **SI-4** (System Monitoring) — same framing as #2017; no numbered BR exists for this observability enhancement on `main` either.

### 7. #2021: cluster-scoped alert mis-correlation in severity triage (main companion of release/v1.5's #2018)

Found by QE against PR #2000's dev-cluster deployment: a `KubePodCrashLooping`-triggered remediation was instead correlated to an unrelated, continuously-firing cluster-wide alert (`CDIDefaultStorageClassDegraded`).

- `severity.Triager.bestAlertMatch` ran two sequential full-fallback passes: a firing-only pass across all three specificity tiers (resource > namespace > cluster), then a pending-only pass only if the firing pass found *nothing at all*. A persistently firing, unrelated cluster-scoped alert would win the firing pass's cluster slot and return immediately — before ever reaching the pending pass, where the target's own resource-specific alert might have been found but simply hadn't started firing yet (a timing race, not a missing alert).
- Replaced the two-pass `bestAlertMatch` with a single-pass `bestOverallMatch`: it scans every alert once, classifies each into a (tier, firing/pending) bucket, and returns the best match ranking **specificity strictly ahead of state** — a merely-pending resource-specific alert now always beats a firing cluster-scoped alert. Within a single tier, firing still beats pending (regression-guarded).
- Split the classification and candidate-update logic into `classifyAlertTier`/`updateTierCandidate` helpers to keep `bestOverallMatch` under `main`'s `gocyclo` threshold (release/v1.5 doesn't enforce this linter yet, so no equivalent split was needed there).
- **BR-HAPI-197**, **FedRAMP SI-4** (accurate correlation of monitoring signals to the actual affected resource).

### 8. #2020: KA finalizes a discovered-but-unanswered decision as a false negative (main companion of release/v1.5's #2019)

Also found by QE against PR #2000's dev-cluster deployment: when `kubernaut_discover_workflows` found and presented a workflow via `kubernaut_present_decision`, but no decision arrived before the interactive session's inactivity timeout, KA finalized the session with a `nil` result — silently discarding a real recommendation as `has_workflow:false`.

- Added `Store.SetPendingDecisionResult`/`Manager.PersistPendingDecisionResult` to preserve the discovered-but-unconfirmed result across the inactivity-timeout/disconnect path (`cmd/kubernautagent/routes.go`'s hardcoded `CompleteUserDriving(id, nil)` calls). `handleDiscoverWorkflows` now calls a new `persistPendingDecision` helper right after a workflow is found (skipped for the `no_matching_workflows` outcome, which is already handled by #2013's `autoCloseOnNoMatchingWorkflows`).
- Threaded a new `decision_expired` `HumanReviewReason` across the full contract: `pkg/kubernautagent/types` constant, KA OpenAPI spec + `agentclient` codegen, DataStorage OpenAPI spec + `ogen-client` codegen, KA's `exactHumanReviewReasons`/`validHumanReviewReasons` whitelists, AA's `mapEnumToSubReason` (→ CRD `SubReason: DecisionExpired`), the `AIAnalysis` CRD's `subReason`/`humanReviewReason` enums, and RO's `mapManualReviewPriority` (routed **High**, not the "nothing found" Medium tier, since a concrete recommendation is waiting for a human).
- Ported from release/v1.5's #2019 fix; the AA-side `SelectedWorkflow` test fixtures were adapted to `main`'s embedded `sharedtypes.WorkflowSnapshot` CRD shape (Issue #1661/DD-WORKFLOW-018), which doesn't exist on release/v1.5.
- **BR-HAPI-197**, **BR-ORCH-036**; **FedRAMP AU-2/AU-3** (accurate audit-of-record — the preserved recommendation and its `decision_expired` reason must be traceable), **AC-6/CM-3** (a presented-but-unanswered decision must never imply approval — `ApprovalRequired` stays false).

## Tests (pyramid invariant)

- `UT-AF-1995-005..007` (#1996 side)
- `UT-AF-1997-001..008` (RED, `phase_guard_aftercallback_1997_test.go`): prove `phaseGuardAfter` must return `(nil, nil)` on both success and failure across every phase-guard path.
- `IT-AF-1997-009` (`phase_guard_afterlog_wiring_1997_test.go`): proves the fix end-to-end through the **real** `llmagent` + `runner.Run` dispatch path, using `root.go`'s actual `AfterToolCallbacks` registration order.
- `UT-KA-2013-001/002` (`discover_workflows_no_match_2013_test.go`): proves the emit-before-complete ordering invariant and a regression guard proving a legitimate recommendation never auto-closes.
- `IT-KA-2013-001` (`discover_workflows_no_match_2013_it_test.go`): drives `InvestigateTool.Handle(action=discover_workflows)` with a **real** `session.Manager` as `autoMgr` (not a mock) and a real `EventLogBridge`, proving the production `EmitSessionEndedByRR` implementation actually propagates `session_ended` to a subscriber — mirrors `terminal_event_1438_it_test.go`'s pattern for the other two call sites.
- `UT-KA-2006-004/005/006` (extends `reasoning_content_stream_1635_test.go`, black-box, drives the real `Investigator.Investigate` entry point): exact-duplicate suppression, whitespace-only-diff suppression, and a regression guard proving genuinely distinct reasoning/narration pairs are unaffected.
- `UT-RO-2012-001/002` (`interactive_timeout_test.go`): reproduce the same-tick Completed/Failed discard bug against the unmodified code (RED), then confirm the `aiTerminal` short-circuit fixes both cases (GREEN); full 9-case pre-existing suite + package regression rerun clean.
- `UT-KA-2016-001/002` (`server_test.go`): prove `WithMCPLogger` forwards SDK-internal log/error events through the `logr` sink, and that a zero-value `MCPDeps.Logger` remains a safe no-op (pre-#2016 default behavior preserved).
- `UT-AF-2021-001..003` (`triage_test.go`): pending resource-specific alert beats firing cluster-scoped; pending namespace-scoped beats firing cluster-scoped; firing still beats pending within the same tier (regression guard).
- `UT-KA-2020-001..010` (`store_test.go`, `discover_workflows_no_match_2013_test.go`, `response_mapper_test.go`, `coverage_668_test.go`): `SetPendingDecisionResult` overwrite/no-op semantics, `discover_workflows` wiring to `PersistPendingDecisionResult` (and its `no_matching_workflows` exemption), `decision_expired` HTTP-response-mapping and audit-whitelist coverage (AU-2/AU-3).
- `UT-AA-2020-006..009` (`response_processor_decision_expired_test.go`): Phase/NeedsHumanReview/HumanReviewReason/SubReason mapping, `SelectedWorkflow` preservation (the actual bug this fixes), and `ApprovalRequired` staying false.
- `IT-AA-2020-001` (`decision_expired_status_test.go`, envtest apiserver CRD acceptance, mirrors the established `IT-AA-1449-001` pattern): proves the API server accepts `decision_expired` in CRD status and the discovered `SelectedWorkflow` persists.
- `UT-RO-2020-001` (`notification_creator_test.go`): `DecisionExpired` SubReason routes to High notification priority.

All new specs RED→GREEN confirmed via `git stash` on the fix commits before writing this description.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/kubernautagent/...` — full suite green, zero regressions
- [x] `go test ./internal/kubernautagent/mcp/tools/... ./internal/kubernautagent/investigator/...` — new specs green (UT-KA-2013-*, IT-KA-2013-001, UT-KA-2006-*)
- [x] `go test ./internal/kubernautagent/mcp/...` — new specs green (UT-KA-2016-001/002), zero regressions
- [x] `go test ./internal/controller/remediationorchestrator/...` — new specs green (UT-RO-2012-001/002), zero regressions
- [x] `go test ./pkg/apifrontend/... ./cmd/apifrontend/...` — full suite green, zero regressions
- [x] `go test ./pkg/aianalysis/... ./pkg/remediationorchestrator/... ./pkg/apifrontend/severity/... ./internal/kubernautagent/... ./pkg/kubernautagent/...` — full suites green, zero regressions
- [x] `go vet ./test/integration/aianalysis/... ./test/integration/kubernautagent/...` — new IT test files compile clean
- [x] `golangci-lint run` on all changed packages — 0 issues (incl. `gocyclo` on the new `bestOverallMatch` split)
- [x] `make manifests` — no unexpected CRD/chart drift beyond the intended `decision_expired`/`DecisionExpired` enum additions
- [x] `make lint-openapi-crd-drift` — pre-existing unrelated `operator_escalation` gap confirmed via `git stash` (not introduced by this PR)
- [x] Confirmed RED→GREEN for all eight fixes on `main`'s diverged code

## Related

- Ports PR #2000 (release/v1.5) to `main`
- Fixes #1996, #1998, #2013, #2006, #2012, #2016, #2021, #2020


jordigilh:
**CI fix**: The `E2E (fleet)` job failed on the previous run with a NodePort collision unrelated to this PR's core fixes:

```
Service "alertmanager-svc" is invalid: spec.ports[0].nodePort: Invalid value: 30193: provided port is already allocated
```

RCA performed — exhaustive static audit found no other claimant of NodePort `30193` anywhere in the `fleet`/`fullpipeline` Kind cluster config, so the exact root cause is unconfirmed (likely a transient dynamic NodePort auto-allocation from an unrelated Service, e.g. Istio/Gateway-API). Relocated `AlertManagerNodePort` 30193 → 30191 (verified free) per [DD-TEST-001 v3.8](https://github.com/jordigilh/kubernaut/blob/main/docs/architecture/decisions/DD-TEST-001-port-allocation-strategy.md), host port 9193 unchanged so no test assertions needed updating.

Opened #2014 to keep a historical record of the incident and its (inconclusive) RCA, so if this recurs on a different port we can tell whether it's a genuinely recurring dynamic-allocation issue and act accordingly.

Pushed in 2493a1b05. Re-running CI now.

PR #2002, opened 2026-08-07
fix(apifrontend): stop phaseGuardAfter short-circuiting the AfterToolCallback chain

jordigilh:
## Summary

- `phaseGuardAfter` re-returned the original `(resp, callErr)` from every exit path instead of `(nil, nil)`. ADK's `invokeAfterToolCallbacks` (`google.golang.org/adk`, `internal/llminternal/base_flow.go`) treats any non-nil result from a callback as an explicit override and stops the chain immediately — so `afterLog` (registered after `afterPhase` in `root.go`'s `AfterToolCallbacks: []llmagent.AfterToolCallback{afterMetrics, afterAudit, afterPhase, afterLog}`) never ran for **any** tool call, silently dropping the `"tool call completed"` audit line.
- This explains the missing completion/timeout-fired logs observed while diagnosing #1995: the tool calls *were* completing, but the audit-trail signal proving it was silently discarded.
- `phaseGuardAfter` only ever records a side effect (session state, the active-context registry) — it never needs to modify `resp`/`callErr` — so `(nil, nil)` is the correct "pass through unchanged" signal, exactly matching how `afterAudit`/`afterMetrics` already behave in the same chain.

## Business requirement / compliance mapping

- **BR-AUDIT-005** (audit completeness) / **FedRAMP AU-3** (content of audit records), **AU-12** (audit generation) — implemented by `afterLog`'s `"tool call completed"` line (`root.go`, TP-1310 §4.3), which this bug was silently dropping for every tool call.

## Tests (pyramid invariant)

- **UT-AF-1997-001..008** (`phase_guard_aftercallback_1997_test.go`, RED): prove `phaseGuardAfter` must return `(nil, nil)` on both success and failure across every phase-guard path (discover_workflows checkpoint, driver-entry, session-terminal, unmatched-tool fallthrough).
- **IT-AF-1997-009** (`phase_guard_afterlog_wiring_1997_test.go`): proves the fix end-to-end through the **real** `llmagent` + `runner.Run` dispatch path, using `root.go`'s actual `AfterToolCallbacks` registration order (`afterPhase` then `afterLog`) — closing the wiring gap that a direct unit call to `phaseGuardAfter` alone cannot observe. Verified this IT test (and all 8 UT cases) fail against the pre-fix code and pass after the fix.

Ref #1997

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/apifrontend/...` (full package suite, zero regressions)
- [x] `go test ./cmd/apifrontend/...`
- [x] `golangci-lint run --timeout=5m ./pkg/apifrontend/agent/...` → 0 issues
- [x] Confirmed RED: reverted the fix locally and re-ran the suite — all 8 UT cases + IT-AF-1997-009 fail as expected
- [x] Confirmed GREEN: re-applied the fix — full suite passes

### DIFF
--- a/pkg/apifrontend/agent/phase_guard.go
+++ b/pkg/apifrontend/agent/phase_guard.go
@@ -160,11 +160,22 @@ func newPhaseGuard(registry *launcher.ActiveContextRegistry) (llmagent.BeforeToo
 			}
 		}
 
+		// #1997: phaseGuardAfter only ever records a side effect (session
+		// state, ActiveContextRegistry) -- it never modifies resp/callErr.
+		// Every exit path below returns (nil, nil), the ADK AfterToolCallback
+		// convention for "pass through unchanged" (google.golang.org/adk,
+		// internal/llminternal/base_flow.go: invokeAfterToolCallbacks treats
+		// a non-nil result as an explicit override that stops the chain,
+		// falling back to the original fResult/fErr only once every
+		// callback has returned nil). Re-returning the original resp/callErr
+		// here looked harmless but silently skipped every AfterToolCallback
+		// registered after this one (afterLog) for any tool call at all,
+		// not just the ones this guard cares about.
 		if !isEntry && !isTerminal && !isDiscoverWorkflows {
-			return resp, callErr
+			return nil, nil
 		}
 		if !isSuccess {
-			return resp, callErr
+			return nil, nil
 		}
 
 		logger := logr.FromContextOrDiscard(ctx)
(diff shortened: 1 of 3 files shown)


jordigilh:
Superseded by #2000 — this commit was cherry-picked onto `fix/1995-workflow-discovery-hang` and is now merged into that PR's diff, so #1995 and #1997 ship together in one PR (fewer PRs to review/merge). See #2000 for the up-to-date description covering both fixes.

PR #2031, opened 2026-08-09
docs: prefer MCP tools (gopls/cocoindex_search) over grep in detection commands

jordigilh:
## Summary
- `AGENTS.md`'s Detection Commands section documented a `gopls MCP > gopls CLI > grep` preference, but never mentioned `cocoindex_search` -- so any search that wasn't an exact-symbol lookup (i.e. semantic/conceptual exploration) had no MCP-preferred path and fell straight to grep.
- Added a `cocoindex_search` subsection alongside the existing `gopls` one, and reframed the preference hierarchy as "MCP tools (gopls/cocoindex_search) > gopls CLI > grep".
- Applied the same MCP-preferred-alternative pattern to the handful of grep-based discovery steps still present in `.cursor/rules/*.mdc` (`03-testing-strategy.mdc`, `04-ai-ml.mdc`, `14-design-decisions-documentation.mdc`). Left literal-pattern/build-log greps alone (e.g. parsing `go build` output, `os.getenv` anti-pattern scan) since there's no MCP equivalent for those.

## Why
This exact correction ("use MCP tools, not grep") has recurred at least twice in agent sessions despite `gopls` guidance already existing in `AGENTS.md`. Root cause: the guidance only covered the Go-symbol-exact case (gopls); the semantic-search case (cocoindex_search) was undocumented, so agents defaulted to grep for that.

## Test plan
- [x] Reviewed rendered `AGENTS.md` Detection Commands section for consistency with existing style
- [x] Confirmed the 3 edited `.mdc` files still parse as valid frontmatter + markdown
- [ ] Next AI agent session in this repo picks `cocoindex_search`/`gopls` over grep for a semantic-search task

### DIFF
--- a/.cursor/rules/03-testing-strategy.mdc
+++ b/.cursor/rules/03-testing-strategy.mdc
@@ -191,18 +191,33 @@ url = config['data_storage']['url']
 
 ### Validation Sequence
 
+Prefer MCP tools over grep for Steps 1-3 (see
+[AGENTS.md Detection Commands](../../AGENTS.md#detection-commands)): gopls gives a type-safe
+definition + method set directly; cocoindex_search finds mocks by meaning when the exact name
+is unknown.
+
 **Step 1 - Interface Discovery**:
+```
+go_search(query="[InterfaceName]")
+```
 ```bash
+# Fallback only if gopls is unavailable:
 grep -r "type.*[InterfaceName].*interface" pkg/ --include="*.go"
 ```
 
 **Step 2 - Method Signature Validation**:
+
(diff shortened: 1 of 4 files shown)


jordigilh:
Closing: this PR's only content was cocoindex_search guidance, which depends on a personal local Engram/Hindsight install (~/.hindsight Postgres + venv) that other contributors don't have. AGENTS.md/CLAUDE.md-equivalent files need to stay reproducible for everyone. Keeping the existing gopls guidance as-is (already universally installable); moving the cocoindex_search preference to a local-only, untracked rule file instead.

Issue #2043, opened 2026-08-09
E2E-FLEET-CC81-001 reconstruction failures have two distinct root causes (AF-created-RR gap, and a recurring #1985 cold-start race)


jordigilh:
## Summary

`test/e2e/fleet/12_reconstruction_test.go`'s `E2E-FLEET-CC81-001` spec has now failed with the identical symptom -- `ReconstructRemediationRequest` returning `400` with `"gateway.signal.received event is required for reconstruction"` -- via **two independently confirmed, structurally different root causes**. Filing this to consolidate both, since they produce the same observable CI failure and were at risk of being conflated.

Related: #1985 (originally filed for root cause #2 below, observed once at the time).

## Root cause #1: AF-created RRs structurally never emit `gateway.signal.received` (new, confirmed via must-gather RCA)

Observed on: [run 31324149042, job 93272706178](https://github.com/jordigilh/kubernaut/actions/runs/31324149042/job/93272706178?pr=2041), must-gather artifact `must-gather-logs-e2e-fleet-31324149042`.

- Failing correlation ID `rr-d8d0d5730e03-11eaf5d3` was never created by Gateway. `datastorage.log` shows the 400; `remediationorchestrator-controller.log:50` shows the RR simply appearing via `"Initializing new RemediationRequest"` with no prior Gateway acceptance. `gateway.log` shows only repeated `"Signal rejected: missing resource info"` for this alert's fingerprint -- Gateway never successfully processed it.
- The actual creator is `test/e2e/fleet/18_af_ka_interactive_fleet_bridge_test.go` (E2E-FLEET-018, issue #1768): its "Turn 1" calls AF's `kubernaut_remediate` A2A/MCP tool, which creates the RR directly via `d.Client.Create()` (`pkg/apifrontend/tools/af_create_rr.go:267`) -- bypassing Gateway's `ProcessSignal`/`emitSignalReceivedAudit` entirely. It correctly emits a different, architecturally valid audit event instead: `audit.EventRRCreated` (`af_create_rr.go:350-362`), never `gateway.signal.received`.
- `MergeAuditData` (`pkg/datastorage/reconstruction/mapper.go:206-216`) hard-requires the literal event type `gateway.signal.received` unconditionally, with no code path for `apifrontend.rr.created`-originated RRs.
[...]

Issue #1985, opened 2026-08-07
flaky(e2e/fleet): Gateway audit events lost during DataStorage cold start — startup-ordering race breaks RemediationRequest reconstruction


jordigilh:
## Summary

Intermittent E2E (fleet) failure: `test/e2e/fleet/12_reconstruction_test.go:89` times out waiting for `ReconstructRemediationRequest` because the `gateway.signal.received` audit event required for reconstruction was never persisted. Root cause is a startup-ordering race between Gateway (and 9 other audit-emitting services) and DataStorage, not a defect in the PR that triggered this run.

Observed on: [jordigilh/kubernaut#1983](https://github.com/jordigilh/kubernaut/pull/1983), [run 31129743441](https://github.com/jordigilh/kubernaut/actions/runs/31129743441/job/92717239570?pr=1983).

## Root cause (must-gather RCA)

- `pkg/audit.BufferedAuditStore` ([pkg/audit/store.go](https://github.com/jordigilh/kubernaut/blob/main/pkg/audit/store.go)) is constructed at process startup by 10 services with **no connectivity gating**: Gateway, AuthWebhook, EffectivenessMonitor, WorkflowExecution, AIAnalysis, SignalProcessing, RemediationOrchestrator, Notification, APIFrontend, KubernautAgent. (DataStorage audits directly to its own Postgres connection and is exempt.)
- On write failure, `writeBatchWithRetry` retries up to `Config.MaxRetries` (default 3, exponential backoff capped at 30s/attempt — roughly 23-90s total) before permanently dropping the batch and logging `"AUDIT DATA LOSS: Dropping batch after max retries"`.
- DataStorage's own cold start (`cmd/datastorage/main.go` Postgres-readiness retry loop, 2s increments) took ~13-20s in the failing run — inside that drop window.
- Gateway's pod started emitting `gateway.signal.received` audit events before DataStorage was reachable. Evidence from must-gather:

  `gateway.log`:
  ```
  {"level":"error","ts":"2026-08-06T23:35:11.470Z","logger":"gateway.server.audit-store","msg":"Failed to write audit batch","attempt":1,"batch_size":2,"error":"network error: ... Client.Timeout exceeded while awaiting headers"}
[...]

PR #2044, opened 2026-08-09
fix(e2e/fleet): make CC8.1 reconstruction test resilient to non-Gateway RRs

jordigilh:
## Summary

- `E2E-FLEET-CC81-001` locked onto the *first* `RemediationRequest` matching a spec-level condition, then retried reconstruction only against that single candidate. This suite has two valid RR-creation paths sharing the same namespace: Gateway's normal webhook ingestion (emits the mandatory `gateway.signal.received` audit event), and E2E-FLEET-018's direct creation via APIFrontend's `kubernaut_remediate` A2A/MCP tool (`pkg/apifrontend/tools/af_create_rr.go`), which emits `apifrontend.rr.created` instead. `pkg/datastorage/reconstruction/mapper.go` hard-requires the former, so if Ginkgo's spec ordering surfaced the APIFrontend-created RR first, reconstruction was **permanently** impossible for it — not a timing issue, so the existing 30s retry never helped.
- Fixed by trying every currently-listed candidate RR on each poll instead of committing to the first one. The test's actual business intent (BR-AUDIT-005 v2.0 CC8.1: cluster_id survives reconstruction for fleet RRs) only needs *one* successfully-reconstructable fleet RR to prove the contract — falling through to the next candidate is correct, not a weaker assertion. This also incidentally tolerates the unrelated, already-tracked #1985 cold-start audit-write race.

Root-caused via must-gather RCA on [run 31324149042](https://github.com/jordigilh/kubernaut/actions/runs/31324149042/job/93272706178?pr=2041). Full RCA (including the deeper reconstruction-mapper gap for `apifrontend.rr.created`-originated RRs, which is NOT fixed here) tracked in #2043.

Test-only change — no production code modified.

## Test plan

- [x] `go build -tags e2e ./test/e2e/fleet/...`
- [x] `go vet -tags e2e ./test/e2e/fleet/...`
- [x] `golangci-lint run --build-tags=e2e ./test/e2e/fleet/...` — 0 issues
- [ ] CI `E2E (fleet)` job green (can't fully reproduce locally without a live fleet Kind cluster; relying on CI for the live-cluster proof)


jordigilh:
Consolidated into #2045, which fixes the actual production root cause (AF-created RRs couldn't be reconstructed at all) and now also includes this PR's E2E test-resilience commits. Closing in favor of #2045 to avoid maintaining two overlapping PRs for the same issue (#2043).

PR #2149, opened 2026-08-14
feat(#2148): auth-only fallback for console-access gate when unconfigured

### LINKED ISSUE #2148: APIFrontend: auth-only fallback for console-access gate when no groups configured
## Problem

AF's coarse-grained \`kubernaut.ai/console\` SAR gate (#1919) is fail-closed with no permissive default: when the operator/Helm chart has no \`consoleAccessGroups\`/\`roleBindings\` configured at all, every authenticated user -- including \`cluster-admin\` -- gets \`403 Forbidden\` on \`GET /a2a/access\` and on every \`/mcp\`/\`/a2a/invoke\` call's inline console check.

This was hit by a real customer deployment (kubernaut-operator side investigation) and is now documented as a hard install prerequisite (kubernaut-operator#334/#335). That's the right floor for anyone who *has* configured RBAC, but for the pure "just got this running, haven't wired OIDC groups yet" case it's an all-or-nothing cliff.

## Proposal

Add an explicit, narrowly-scoped auth-only fallback for the **console gate only** (not the per-tool SAR checks, which remain unconditionally fail-closed):

- New \`RBACConfig.ConsoleAccessAuthOnly bool\` (default \`false\`, i.e. today's behavior unchanged).
- When \`true\`, \`SARChecker.CheckConsoleAccess\` skips the SAR call entirely and returns \`(true, nil)\` for any non-empty authenticated user -- authentication is still mandatory (empty user still fails closed), only the authorization SAR step is skipped.
- \`Check\` (per-tool authorization) is completely unaffected -- always requires SAR, always fail-closed. No change to that invariant.
- kubernaut-operator sets this flag \`true\` only when \`spec.apiFrontend.rbac\` is entirely unset (nil) -- i.e. the installer has never touched RBAC config at all. Any deployment with even one \`roleBindings\` entry, or an explicit \`consoleAccessGroups\` (including explicit \`[]\` opt-out), is unaffected and keeps today's exact behavior.

This intentionally narrows the scope of a prior decision that AF must always require SAR fail-closed with no dev-mode bypass -- that invariant is preserved in full for the per-tool gate; this PR adds one narrowly-scoped, explicitly-configured exception for the advisory/coarse-grained console gate only, which by its own design docs was never itself the primary security boundary (per-tool RBAC is).

## Companion change

kubernaut-operator PR to follow, wiring \`afRBACConfig\` to set this flag based on \`spec.apiFrontend.rbac == nil\`.

jordigilh:
## Summary

Closes #2148.

- AF's coarse-grained `kubernaut.ai/console` SAR gate (#1919) is fail-closed with no permissive default: when no `consoleAccessGroups`/`roleBindings` are configured at all, every authenticated user (including `cluster-admin`) gets `403` on the Console and on every `/mcp`/`/a2a/invoke` call's inline console check, with zero fallback. This was hit by a real customer deployment and is now a documented hard install prerequisite (kubernaut-operator#334/#335) -- correct for anyone who *has* configured RBAC, but an all-or-nothing cliff otherwise.
- Adds `RBACConfig.ConsoleAccessAuthOnly` (default `false`, today's behavior unchanged). When `true`, `SARChecker.CheckConsoleAccess` grants any authenticated (non-empty) user without calling SAR -- authentication stays mandatory, only the group-based authorization step is skipped.
- `Check` (per-tool authorization) is **completely unaffected** -- always calls SAR, always fail-closed, no bypass. This narrows the scope of the existing "AF must always require SAR, no dev-mode fallback" invariant to exclude only the advisory, coarse-grained console gate, which per its own design docs was never itself the primary security boundary (per-tool RBAC is).
- A companion `kubernaut-operator` PR will set this flag based on `spec.apiFrontend.rbac == nil` (i.e. RBAC never configured at all -- any deployment with even one `roleBindings` entry or an explicit `consoleAccessGroups` is unaffected).

## Test plan
- [x] `UT-AF-2148-001`: auth-only mode grants any authenticated user without calling SAR.
- [x] `UT-AF-2148-002`: auth-only mode still fail-closed on empty user (authentication remains mandatory).
- [x] `UT-AF-2148-003`: per-tool `Check` is unaffected by the flag -- still calls SAR, still fail-closed.
- [x] Full `pkg/apifrontend/...` + `cmd/apifrontend/...` suite passes.
- [x] `go vet` clean.

### DIFF
--- a/cmd/apifrontend/main.go
+++ b/cmd/apifrontend/main.go
@@ -52,7 +52,7 @@ func run() int {
 	defer func() { _ = zapLogger.Sync() }()
 	ctrl.SetLogger(logger.WithName("controller-runtime"))
 
-	sarChecker, err := buildSARClient(logger, cfg.RBAC.SARCacheTTL)
+	sarChecker, err := buildSARClient(logger, cfg.RBAC.SARCacheTTL, cfg.RBAC.ConsoleAccessAuthOnly)
 	if err != nil {
 		return 1
 	}
@@ -387,8 +387,11 @@ func setupConfigAndLogger() (*config.Config, logr.Logger, *zap.Logger, error) {
 }
 
 // buildSARClient wires the SelfSubjectAccessReview-based authorizer used for
-// K8s-tool RBAC gating.
-func buildSARClient(logger logr.Logger, sarCacheTTL time.Duration) (auth.ToolAuthorizer, error) {
+// K8s-tool RBAC gating. consoleAccessAuthOnly configures the coarse-grained
+// console gate (only) to skip its SAR call and grant any authenticated user
+// -- see RBACConfig.ConsoleAccessAuthOnly (#2148); per-tool checks are
+// unaffected and always call SAR.
+func buildSARClient(logger logr.Logger, sarCacheTTL time.Duration, consoleAccessAuthOnly bool) (auth.ToolAuthorizer, error) {
 	restCfg, err := ctrl.GetConfig()
 	if err != nil {
 		logger.Error(err, "failed to get in-cluster config for SAR client")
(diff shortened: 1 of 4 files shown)


jordigilh:
Closing in favor of proposal-only: #2148 already captures the design; this PR jumped ahead to an implementation before upstream (AF) maintainers weighed in, which wasn't the intent. Leaving the `feat/console-access-auth-only-fallback` branch/commits in place as an optional reference implementation if it's useful when #2148 is discussed -- not asking for it to be merged as-is.

Issue #2221, opened 2026-08-21
chart: mcp.toolTimeout / mcp.toolTimeouts never exposed via schema or template


jordigilh:
## Problem

AF's `MCPConfig` has two real, actively-consumed fields beyond `sessionIdleTimeout`:

```go
// pkg/apifrontend/config/config.go:242-248
type MCPConfig struct {
    Enabled            bool
    SessionIdleTimeout time.Duration `yaml:"sessionIdleTimeout,omitempty"`
    ToolTimeout        time.Duration `yaml:"toolTimeout,omitempty"`
    ToolTimeouts       map[string]time.Duration `yaml:"toolTimeouts,omitempty"`
}
```

Both are wired into the MCP bridge (`cmd/apifrontend/mcp_a2a_handlers.go:44-45`) and have real built-in defaults in `config.Default()`:

```go
ToolTimeout: 30 * time.Second,
ToolTimeouts: map[string]time.Duration{
    "kubernaut_investigate":        15 * time.Minute,
    "kubernaut_await_session":      3 * time.Minute,
    "kubernaut_watch":              15 * time.Minute,
    "kubernaut_discover_workflows": 60 * time.Second,
},
```

But `values.schema.json` has no `mcp.toolTimeout`/`mcp.toolTimeouts` entries, and `templates/apifrontend/apifrontend.yaml`'s `mcp:` block only emits `enabled` and `sessionIdleTimeout` (lines 95-96). A chart-only install has no way to override per-tool MCP timeouts at all, and — separately — the chart's rendered ConfigMap never sets these fields explicitly, relying entirely on the binary's Go-level defaults rather than making them visible/auditable in the deployed config.

## Proposed fix

Add `apifrontend.config.mcp.toolTimeout` and `apifrontend.config.mcp.toolTimeouts` (map of tool name → duration) to `values.schema.json` with the four defaults above, and render them explicitly in `templates/apifrontend/apifrontend.yaml`'s `mcp:` block.

## Context

Found during a cross-service config-knob consistency audit for `kubernaut-operator#173`. `kubernaut-operator` independently sets these fields in its ConfigMap generator but was missing 2 of the 4 upstream per-tool overrides (fixed in `kubernaut-operator#374`) — this chart gap is exactly why that drift went unnoticed for as long as it did: there's no schema to diff against.

jordigilh:
## Follow-up from kubernaut-operator#374: recommend re-scoping, not the "Proposed fix" above

[...]

Issue #2220, opened 2026-08-21
chart bug: mcp.sessionIdleTimeout schema default (5m) is stale vs. AF binary's actual built-in default (30m)


jordigilh:
## Problem

`values.schema.json` declares `apifrontend.config.mcp.sessionIdleTimeout` with `"default": "5m"`.

But the AF binary itself hardcodes a different fallback:

```go
// cmd/apifrontend/mcp_a2a_handlers.go:56-58
mcpSessionTimeout := d.Cfg.MCP.SessionIdleTimeout
if mcpSessionTimeout == 0 {
    mcpSessionTimeout = 30 * time.Minute
}
```

So a fresh `helm install` with no override for `mcp.sessionIdleTimeout` renders `5m` into the ConfigMap — which *does* get applied (unlike #2219), so this isn't a silent-no-op bug, but it means the chart's stated default has drifted from the binary's own intended default and no longer documents real behavior accurately. Anyone reading `values.schema.json` to understand default behavior gets the wrong number.

## Proposed fix

Bump the schema default for `apifrontend.config.mcp.sessionIdleTimeout` from `5m` to `30m` to match `cmd/apifrontend/mcp_a2a_handlers.go`, or (preferred) remove the Go-side fallback and always require the chart/operator to populate a real value explicitly, so there's a single source of truth.

## Context

Found during a cross-service config-knob consistency audit for `kubernaut-operator#173`. `kubernaut-operator` independently hardcodes `30m` for this field, which — per this issue — is the value that actually matches production behavior; it's the chart that needs to move, not the operator.

jordigilh:
Fixed by PR #2225 (merged 2026-08-21T15:03:39Z, commit `c1740830e`) — see #2217's closing comment for the shared multi-issue PR/auto-close note. Verified on `origin/main`: `mcp.sessionIdleTimeout` schema default consolidated to `30m` (matching the Go-side fallback that was actually being applied), single source of truth in `config.DefaultConfig()`.

Issue #2219, opened 2026-08-21
chart bug: DataStorage server.readTimeout schema override is silently ignored; writeTimeout has no schema field at all


jordigilh:
## Problem

`values.schema.json` declares `datastorage.config.server.readTimeout` as a configurable `goDuration` field (default `30s`, description "Server read timeout"), so `helm install --set datastorage.config.server.readTimeout=60s` validates cleanly against the schema.

But `templates/datastorage/datastorage.yaml:19-20` renders both `readTimeout` and `writeTimeout` as bare literals, never referencing `.Values`:

```yaml
readTimeout: 30s
writeTimeout: 30s
```

**Impact:** setting `datastorage.config.server.readTimeout` via `--set` or `values.yaml` passes schema validation but has **no effect on the deployed ConfigMap** — the override is silently discarded. This is worse than a missing knob: it looks configurable and fails silently.

`writeTimeout` additionally has no schema field at all, despite being an exact peer of the schema-declared `readTimeout`.

## Proposed fix

1. Update the template to reference `{{ $v.config.server.readTimeout }}` / `{{ $v.config.server.writeTimeout }}` instead of the literals.
2. Add a `writeTimeout` schema field alongside `readTimeout` (same shape, default `30s`).

## Context

Found during a cross-service config-knob consistency audit for `kubernaut-operator#173`. `kubernaut-operator` independently hardcodes the same `30s`/`30s` pair in its ConfigMap generator, consistent with the chart's current (broken) behavior — worth keeping in sync once this is fixed upstream.

jordigilh:
Fixed by PR #2225 (merged 2026-08-21T15:03:39Z, commit `f6b0308d2`) — see #2217's closing comment for the shared multi-issue PR/auto-close note. Verified on `origin/main`: `datastorage.server.readTimeout` override is no longer silently discarded, and `writeTimeout` now has its own schema field (both defaults `30s` preserved).

Issue #2218, opened 2026-08-21
chart: DataStorage database.connMaxIdleTime hardcoded while its 3 struct siblings are schema-driven


jordigilh:
## Problem

`templates/datastorage/datastorage.yaml` hardcodes `database.connMaxIdleTime`:

```yaml
# templates/datastorage/datastorage.yaml:43-46
maxOpenConns: {{ $v.config.database.maxOpenConns }}
maxIdleConns: {{ $v.config.database.maxIdleConns }}
connMaxLifetime: {{ $v.config.database.connMaxLifetime }}
connMaxIdleTime: 10m
```

`maxOpenConns`, `maxIdleConns`, and `connMaxLifetime` — all three siblings in the same `database` config struct — are schema-driven (`values.schema.json` `datastorage.config.database.*`, defaults 100/20/`1h`). `connMaxIdleTime` alone is a bare literal with no schema entry.

## Proposed fix

Add `datastorage.config.database.connMaxIdleTime` to `values.schema.json` (`goDuration`, default `10m` to preserve current behavior), and reference it from the template.

## Context

Found during a cross-service config-knob consistency audit for `kubernaut-operator#173`.

jordigilh:
Fixed by PR #2225 (merged 2026-08-21T15:03:39Z, commit `b9e2a3c2b`) — see #2217's closing comment for the shared multi-issue PR/auto-close note. Verified on `origin/main`: `datastorage.database.connMaxIdleTime` is now schema-driven (default `10m` preserved).

PR #2225, opened 2026-08-21
fix(chart): close 6 values.schema.json config-knob gaps (#2216-#2221)

### LINKED ISSUE #2216: chart: SignalProcessing enrichment.cacheTtl / classifier.hotReloadInterval hardcoded, no schema field
## Problem

`templates/signalprocessing/signalprocessing.yaml` hardcodes two knobs as bare literals, with no corresponding `values.schema.json` entry:

```yaml
# templates/signalprocessing/signalprocessing.yaml:64-70
enrichment:
  cacheTtl: "5m"
classifier:
  hotReloadInterval: "30s"
```

Neither can be overridden via `helm install --set`.

## Why this is inconsistent

Cache-TTL knobs already have an established, schema-driven pattern elsewhere in this same chart:
- API Frontend `severityTriage.cacheTTLSeconds` (schema-driven, default `30`)
- API Frontend `rbac.sarCacheTTL` (schema-driven, default `30s`)

`enrichment.cacheTtl` is the same knob category (a cache lifetime) but was never exposed. `classifier.hotReloadInterval` is the same category as the interval/backoff knobs RemediationOrchestrator already exposes in full (`routing.exponentialBackoffBase`, `scopeBackoffBase`, etc.).

## Proposed fix

Add `signalprocessing.config.enrichment.cacheTtl` and `signalprocessing.config.classifier.hotReloadInterval` to `values.schema.json` (`goDuration`, defaults `5m`/`30s` to preserve current behavior), and reference them from the template.

## Context

Found during a cross-service config-knob consistency audit for `kubernaut-operator#173`.

jordigilh:
## Summary

Closes a batch of chart/values.schema.json config-knob gaps found during a cross-service consistency audit for `kubernaut-operator#173`. Each issue is its own commit, RED→GREEN via a dedicated `helm-unittest` suite.

- #2216: signalprocessing `enrichment.cacheTtl` / `classifier.hotReloadInterval` were bare template literals despite being actively-read Go config fields (defaults `5m`/`30s` preserved).
- #2217: gateway `server.idleTimeout` was a bare literal while its 3 struct siblings were schema-driven (default `120s` preserved).
- #2218: datastorage `database.connMaxIdleTime` was a bare literal while its 3 struct siblings were schema-driven (default `10m` preserved).
- #2219: datastorage `server.readTimeout` was schema-declared but the override was **silently discarded** by the template (real bug, not just a missing knob); `writeTimeout` had no schema field at all despite being an exact peer. Both fixed; defaults preserved (`30s`/`30s`).
- #2220: apifrontend `mcp.sessionIdleTimeout` schema default was `5m`, inconsistent with the `30m` the Go-side `cmd/apifrontend` fallback silently applied whenever the field was unset — so a bare install actually got a real 5m timeout, not the 30m an operator reading the Go source would expect. Consolidated to a single 30m source of truth: schema bumped to `30m`, `config.DefaultConfig()` now sets it explicitly, and the now-redundant `cmd/apifrontend` fallback removed.
- #2221: apifrontend `mcp.toolTimeout` / `mcp.toolTimeouts` were wired into the MCP bridge with real Go-side defaults but never exposed via the chart at all. `toolTimeouts` is modeled as 4 named schema properties (not an arbitrary map) so each of the 4 known tool defaults materializes correctly and a partial override of one tool can never wipe out the other 3 (proven empirically by the new test).

**#2215 was investigated and found obsolete** — its premise (hardcoded `aianalysis.agent.timeout`/`sessionPollInterval`) was removed as dead code by #2204 the day before #2215 was filed. Commented on the issue recommending closure; not included in this PR.

Each fix follows RED→GREEN: a new `helm-unittest` suite proving the gap/bug first (asserting the exact Go-side default and, where relevant, an explicit regression-guard proving an override that was previously silently discarded now renders through), then the minimal schema/template/Go change to make it pass. Full `helm unittest` suite (563 tests, 68 suites) is green throughout, `go build ./...` and `golangci-lint` are clean.

## Commits

1. `afb2ea316` fix(chart): expose signalprocessing enrichment.cacheTtl / classifier.hotReloadInterval (#2216)
2. `f030e7897` fix(chart): expose gateway server.idleTimeout in values.schema.json (#2217)
3. `b9e2a3c2b` fix(chart): expose datastorage database.connMaxIdleTime in values.schema.json (#2218)
4. `f6b0308d2` fix(chart): datastorage server.readTimeout override was silently discarded; add writeTimeout schema field (#2219)
5. `c1740830e` fix(chart,apifrontend): consolidate mcp.sessionIdleTimeout default to 30m, single source of truth (#2220)
6. `2262c3db8` fix(chart): expose apifrontend mcp.toolTimeout / mcp.toolTimeouts in values.schema.json (#2221)
7. `1daa131de` refactor(chart): fix stale mcp.sessionIdleTimeout example default, document toolTimeout/toolTimeouts

## Test plan

- [x] New `helm-unittest` suite per issue, RED-confirmed against pre-fix state, GREEN after fix
- [x] Full `helm unittest charts/kubernaut/` suite green (68 suites / 563 tests)
- [x] `go build ./...` clean
- [x] `golangci-lint run` clean on all touched Go packages
- [x] `make generate-helm-defaults generate-helm-config-docs` re-run and committed after every schema change

Closes #2216, #2217, #2218, #2219, #2220, #2221

### DIFF
--- a/charts/kubernaut/templates/_generated_defaults.tpl
+++ b/charts/kubernaut/templates/_generated_defaults.tpl
@@ -135,6 +135,7 @@ datastorage:
             existingSecret: ""
             secretKey: hmacKey
         database:
+            connMaxIdleTime: 10m
             connMaxLifetime: 1h
             maxIdleConns: 20
             maxOpenConns: 100
--- a/charts/kubernaut/templates/_generated_defaults.tpl
+++ b/charts/kubernaut/templates/_generated_defaults.tpl
@@ -50,6 +50,12 @@ apifrontend:
         mcp:
             enabled: true
             sessionIdleTimeout: 30m
+            toolTimeout: 30s
+            toolTimeouts:
+                kubernaut_await_session: 3m
+                kubernaut_discover_workflows: 60s
+                kubernaut_investigate: 15m
+                kubernaut_watch: 15m
         rateLimit:
             ipRequestsPerSec: 10000
             maxConcurrentSessions: 50
--- a/charts/kubernaut/templates/_generated_defaults.tpl
+++ b/charts/kubernaut/templates/_generated_defaults.tpl
@@ -434,6 +434,11 @@ remediationorchestrator:
         enabled: true
     replicas: 1
 signalprocessing:
+    config:
+        classifier:
+            hotReloadInterval: 30s
+        enrichment:
+            cacheTtl: 5m
     fleet:
         namespace: ""
         oauth2:
--- a/charts/kubernaut/templates/_generated_defaults.tpl
+++ b/charts/kubernaut/templates/_generated_defaults.tpl
@@ -49,7 +49,7 @@ apifrontend:
             enabled: true
         mcp:
             enabled: true
-            sessionIdleTimeout: 5m
+            sessionIdleTimeout: 30m
         rateLimit:
             ipRequestsPerSec: 10000
             maxConcurrentSessions: 50
--- a/charts/kubernaut/templates/_generated_defaults.tpl
+++ b/charts/kubernaut/templates/_generated_defaults.tpl
@@ -215,6 +215,7 @@ gateway:
         middleware:
             trustedProxyCIDRs: []
         server:
+            idleTimeout: 120s
             k8sRequestTimeout: 15s
             maxConcurrentRequests: 100
             readTimeout: 30s
--- a/charts/kubernaut/templates/_generated_defaults.tpl
+++ b/charts/kubernaut/templates/_generated_defaults.tpl
@@ -160,6 +160,7 @@ datastorage:
                 requestsPerSecond: 50
             readTimeout: 30s
             signerCertDir: /etc/certs
+            writeTimeout: 30s
     dbExistingSecret: ""
     logging:
         level: INFO
--- a/charts/kubernaut/templates/apifrontend/apifrontend.yaml
+++ b/charts/kubernaut/templates/apifrontend/apifrontend.yaml
@@ -94,6 +94,15 @@ session:
 mcp:
   enabled: {{ $v.config.mcp.enabled }}
   sessionIdleTimeout: {{ $v.config.mcp.sessionIdleTimeout | quote }}
+  # Issue #2221: toolTimeout/toolTimeouts were wired into the MCP bridge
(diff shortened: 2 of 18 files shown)


jordigilh:
Nice work closing out 6 of the 7 `#173` consistency-audit issues in one pass — RED→GREEN per field with dedicated `helm-unittest` suites is exactly the rigor this class of bug needs.

One thing worth flagging for a possible lightweight follow-up: for the fields that had a real Go-side default already (`mcp.sessionIdleTimeout`, `mcp.toolTimeout`, `mcp.toolTimeouts`), this PR now maintains that default in two independently-authored places — `pkg/apifrontend/config/config.go`'s `DefaultConfig()` (Go literal) and `charts/kubernaut/values.schema.json`'s `"default"` (JSON literal), synced only by `make generate-helm-defaults` + code review. That's structurally the same *shape* of risk that caused `kubernaut-operator#374`: two hardcoded copies of one conceptual value that can silently diverge if only one side is updated (e.g. a future change to `DefaultConfig()`'s `ToolTimeouts` map without a matching schema-default bump).

`TC-P2C-03d` here is a good guard for *this* field, but it only proves the Go side is internally consistent — it doesn't cross-check against the schema literal. A generic version of that guard (parse `values.schema.json`'s `default` for each `config.DefaultConfig()`-backed field, assert it round-trips to the same `time.Duration`/value as the Go default) would catch drift in either direction for all such fields at once, rather than needing a new one-off test every time a new dual-sourced default is added.

Not a blocker for this PR — just flagging it as a possible fast-follow hardening issue if useful. Happy to file it separately if so.

jordigilh:
Folded this into the stacked follow-up: #2227 adds a generic, table-driven `TC-P2C-04` guard (`pkg/apifrontend/config/schema_default_drift_test.go`) that cross-checks `config.DefaultConfig()`'s Go value against `values.schema.json`'s declared default for every dual-sourced MCP duration field (`sessionIdleTimeout`/`toolTimeout`/`toolTimeouts.*`) -- closing the same drift class as kubernaut-operator#374. Adding a future dual-sourced default only needs a new `Entry()` row.

Issue #1390, opened 2026-06-09
Bug: Duplicate investigation submission causes stuck Analyzing phase (ghost investigation race)


jordigilh:
## Summary

A remediation request can get permanently stuck in the `Analyzing` phase when the investigation handler accepts a duplicate submission for the same incident. The duplicate spawns a "ghost" non-interactive investigation that races with the original, preventing the normal RCA → workflow selection → session release lifecycle from completing.

## Observed Behavior

`rr-ea7d9b451b31-14926515` (KubePodCrashLooping on `demo-webui/web-frontend`) entered `Analyzing` at 18:27:43Z and never progressed. The AIAnalysis controller polled indefinitely every 15s with `"Session under user control, continuing to poll"` but empty `progress` and empty `actingUser`.

## Root Cause

The investigation was submitted **twice** to the KA handler (`server/handler.go:111`) for the same `incident_id` within 4 seconds (18:27:43 and 18:27:47):

```
{"msg":"investigation submitted","incident_id":"ai-rr-ea7d9b451b31-14926515","signal":"KubePodCrashLooping",...}  // 18:27:43
{"msg":"investigation submitted","incident_id":"ai-rr-ea7d9b451b31-14926515","signal":"KubePodCrashLooping",...}  // 18:27:47
```

### What happened

1. **First submission** (18:27:43): Spawned the interactive investigation correctly — session `e3f309a3` created, MCP session `4dbc3176` wired, streaming sink active. RCA ran 8 turns, completed at 18:29:09 (`sentinel detected`, `submit_result`, phase `rca`).

2. **Second submission** (18:27:47): Spawned a **duplicate non-interactive investigation** — `session_id: ""`, `sink is nil`, falling back to non-streaming Chat. This ghost investigation re-ran the entire RCA phase independently.

3. **Ghost investigation anomaly** (18:30:12): The duplicate hit the anomaly detector's per-tool call limit: `execute_prometheus_instant_query` exceeded limit (11 > 10). The LLM was looping on Prometheus queries without session context.

4. **No workflow selection** (18:30:37+): After both RCA sentinels fired, `runWorkflowSelection` was never invoked. Compare with the successful prior investigation (`rr-618ac7d3b894`) where the sequence was: RCA sentinel → `runWorkflowSelection` → `submit_result_with_workflow` → session released (`reason: workflow_selected`).

[...]

PR #2234, opened 2026-08-21
fix(kubernautagent): synthesize curated placeholder for nil-result session completion

### LINKED ISSUE #2233: kubernaut-agent panics (nil pointer) when a user-driving MCP session disconnects before producing any result
## Summary

`kubernaut-agent` crashes with a nil-pointer panic (SIGSEGV) whenever a user-driving MCP session's client disconnects (grace period expires) *before the investigation ever produced a result*. The crash takes down the whole process, causing every other concurrent MCP session on that pod to see `connection reset by peer` until the pod restarts.

Discovered while triaging [CI run 32524002195](https://github.com/jordigilh/kubernaut/actions/runs/32524002195/job/96905096732) (`E2E (fullpipeline)` on `main`, post-merge of #2229): `E2E-FP-1456-001` failed with `MCP connect: calling "initialize": ... connection reset by peer` because the KA pod was mid-crash-restart (`describe.txt`: `Last State: Terminated, Reason: Error, Exit Code: 2, Restart Count: 2`) -- caused by a *different*, unrelated session's disconnect.

## Panic (from must-gather `previous.log`)

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x1e7b24f]

goroutine 1736 [running]:
github.com/jordigilh/kubernaut/internal/kubernautagent/agentsession.buildRootCauseAnalysisMap(0x0)
	internal/kubernautagent/agentsession/mapping.go:150
github.com/jordigilh/kubernaut/internal/kubernautagent/agentsession.MapInvestigationResultToAgentSessionResult(...)
	internal/kubernautagent/agentsession/mapping.go:107
github.com/jordigilh/kubernaut/internal/kubernautagent/agentsession.(*Dispatcher).writeTerminalStatus.func2(...)
	internal/kubernautagent/agentsession/status_writer.go:168
github.com/jordigilh/kubernaut/internal/kubernautagent/agentsession.(*Dispatcher).updateStatus(...)
	internal/kubernautagent/agentsession/status_writer.go:59
github.com/jordigilh/kubernaut/internal/kubernautagent/agentsession.(*Dispatcher).writeTerminalStatus(...)
	internal/kubernautagent/agentsession/status_writer.go:154
github.com/jordigilh/kubernaut/internal/kubernautagent/agentsession.(*Dispatcher).OnTerminal(...)
	internal/kubernautagent/agentsession/dispatcher.go:573
github.com/jordigilh/kubernaut/internal/kubernautagent/session.(*Manager).fireTerminalHook(...)
	internal/kubernautagent/session/manager.go:117
github.com/jordigilh/kubernaut/internal/kubernautagent/session.(*Manager).CompleteUserDriving(...)
	internal/kubernautagent/session/manager_query.go:359
github.com/jordigilh/kubernaut/internal/kubernautagent/mcp/tools.CompleteHTTPSession(...)
	internal/kubernautagent/mcp/tools/session_teardown.go:52
main.buildMCPCoreDeps.buildMCPDisconnectHandler.func4(...)
	cmd/kubernautagent/routes.go:787
github.com/jordigilh/kubernaut/internal/kubernautagent/mcp.(*GracefulSessionClosedHandler).Run.func1(...)
	internal/kubernautagent/mcp/disconnect_handler.go:134
```

## Root cause

1. The MCP disconnect handler calls `CompleteHTTPSession(..., result: nil, ...)` -> `Manager.CompleteUserDriving(id, nil)` after a session's grace period expires (`internal/kubernautagent/mcp/tools/session_teardown.go:52`).
2. `CompleteUserDriving` (`internal/kubernautagent/session/manager_query.go:302-362`) reads `finalResult = sess.Result` under lock. If the investigation never produced a result before the disconnect, `sess.Result` is also `nil`, so `finalResult` stays `nil`.
3. It unconditionally calls `fireTerminalHook(id, correlationID, StatusCompleted, finalResult /* nil */, nil)` -- there's no branch for "disconnected with no result ever produced."
4. `writeTerminalStatus`'s `default` case (`session.StatusCompleted`) unconditionally calls `MapInvestigationResultToAgentSessionResult(log, result, incidentID)` with `result == nil` (`internal/kubernautagent/agentsession/status_writer.go:143-169`).
5. `MapInvestigationResultToAgentSessionResult`/`buildRootCauseAnalysisMap` (`internal/kubernautagent/agentsession/mapping.go:101-150`) dereference `r.RCASummary` etc. with `r == nil` -> panic.

This function has never had a nil-guard since it was introduced in #2170 (2026-08-18, the AA<->KA AgentSession-CRD migration) -- pre-existing latent bug, unrelated to #2229 (ADK v2.0.0 migration); the CI run just happened to trip it.

## Impact

Any MCP client disconnecting from an interactive/user-driving investigation before it produces a result (e.g. user closes their client early, or an E2E/production timeout) crashes the entire `kubernaut-agent` process, disrupting every other concurrent investigation on that pod until it restarts. This is an availability/reliability bug, not merely a test flake.

## Proposed fix

- In `CompleteUserDriving`, when `finalResult == nil`, fire the terminal hook with `StatusFailed` (curated message, reusing the existing `curatedFailureMessage` pattern) instead of blindly `StatusCompleted`.
- Defense-in-depth: make `MapInvestigationResultToAgentSessionResult`/`buildRootCauseAnalysisMap` nil-safe, consistent with this same file's `marshalJSON` which already documents a "never panics on malformed result" invariant that this function currently violates.

## Related

- Introduced alongside #2170 (AA<->KA AgentSession CRD migration)
- Observed on CI run for #2229 (unrelated ADK v2.0.0 migration -- pod crash was caused by a concurrent, unrelated session's disconnect)

jordigilh:
## Summary

- Fixes #2233: `kubernaut-agent` panicked (nil pointer dereference) whenever a user-driving MCP session reached `StatusCompleted` with no `InvestigationResult` ever stored -- e.g. the MCP client disconnects, or the session's own inactivity timeout fires, before an investigation produces a result. This crash-looped the pod and dropped every other in-flight MCP session on it (root-caused from CI run [32524002195](https://github.com/jordigilh/kubernaut/actions/runs/32524002195/job/96905096732), `E2E-FP-1456-001`, via must-gather `previous.log`).
- `writeTerminalStatus`'s `StatusCompleted` branch calls `MapInvestigationResultToAgentSessionResult` unconditionally; a nil `InvestigationResult` dereferenced inside `buildRootCauseAnalysisMap`.
- Fix mirrors existing precedent rather than inventing a new outcome: the retired HTTP-polling handler's `synthesizeNilResult` default branch (#1390, `internal/kubernautagent/server/handler.go`) already produces `"Investigation completed without result"` / `Confidence: 0` for this exact class of event, and AA's `isSessionTimedOutWithoutResult` (`pkg/aianalysis/handlers/investigating.go`) already special-cases that literal to reclassify `SubReason=InvestigationInconclusive` -- so the CRD-native mapping function now synthesizes the identical placeholder when `r == nil`, and **no AA-side change is required**.

## Note on scope vs. earlier discussion

An earlier plan for this fix (before investigating this deeply) proposed marking the session `StatusFailed` with a curated message when the result is nil. Preflight for this PR surfaced the `synthesizeNilResult`/`isSessionTimedOutWithoutResult` precedent above, which is a materially better, already-precedented, already-AA-compatible fix -- so I went with reusing that pattern instead of introducing a new `StatusFailed` branch, which would have been inconsistent with how KA already handles the identical scenario on its legacy HTTP-polling path.

## Test plan

- `UT-AA-KA-065-016` (`internal/kubernautagent/agentsession/mapping_test.go`): pins the nil-guard directly at the mapping-function boundary -- asserts no panic and the exact placeholder shape AA expects.
- `IT-AA-KA-2233-001` (`internal/kubernautagent/agentsession/dispatcher_test.go`): reproduces the crash end-to-end through the real production wiring (`session.Manager` -> `Dispatcher` -> fake `AgentSession` client) -- confirmed this test panics against the pre-fix code (RED) and passes against the fix (GREEN).
- `go build ./...`, `golangci-lint run ./internal/kubernautagent/agentsession/...` (0 issues), `go test ./internal/kubernautagent/... ./pkg/aianalysis/...` (all green, no regressions).

- [x] `go build ./...`
- [x] `golangci-lint run`
- [x] Unit test (mapping-function nil-guard)
- [x] Integration test (full wiring, reproduces the original crash)
- [x] No AA-side changes needed or made

### DIFF
--- a/internal/kubernautagent/agentsession/dispatcher_test.go
+++ b/internal/kubernautagent/agentsession/dispatcher_test.go
@@ -47,6 +47,7 @@ import (
 type fakeInvestigationRunner struct {
 	calls  atomic.Int32
 	delay  time.Duration
+	block  chan struct{}
 	result *katypes.InvestigationResult
 	err    error
 }
@@ -56,6 +57,18 @@ func (f *fakeInvestigationRunner) Investigate(ctx context.Context, _ katypes.Sig
 	if f.delay > 0 {
 		select {
 		case <-time.After(f.delay):
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+	}
+	// block, when set, lets a test hold this goroutine "running" (never
+	// producing a result) until it explicitly releases it or the context
+	// is cancelled -- used to reproduce a user-driving session whose
+	// client disconnects before any InvestigationResult is ever stored
+	// (issue #2233).
+	if f.block != nil {
+		select {
+		case <-f.block:
 		case <-ctx.Done():
 			return nil, ctx.Err()
 		}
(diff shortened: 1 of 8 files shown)


jordigilh:
Rebased onto \`main\` to resolve a merge conflict: [commit 76bc20270](https://github.com/jordigilh/kubernaut/commit/76bc20270) (merged via #2222, same day) independently added a nil-guard to the same function for the same underlying bug, using a different, less precise curated literal (\`"session completed with no investigation result recorded"\`) that doesn't match what AA's \`isSessionTimedOutWithoutResult\` already special-cases. Resolved by superseding that guard's content with this PR's placeholder (\`"Investigation completed without result"\` / \`Confidence: 0\`), keeping its \`log.Info\` observability call, and updating \`UT-AA-KA-065-016\` accordingly. Verified \`go build ./...\`, \`golangci-lint\`, and the full \`internal/kubernautagent/...\` + \`pkg/aianalysis/...\` suites all green post-rebase.

Also investigated the \`Integration (remediationorchestrator)\` CI failure on the pre-rebase run ([32529705863](https://github.com/jordigilh/kubernaut/actions/runs/32529705863/job/96923476496)) via its must-gather artifact: unrelated to this PR's diff. It's a pre-existing, narrow resourceVersion race between the RAR approval test's unguarded \`Status().Update()\` and \`RARReconciler\`'s own retry-protected audit-condition write. Filed #2235 with full RCA and a suggested fix; not addressed here since it's out of scope for this PR's files.

jordigilh:
Added a second, separate fix to this branch: #2235 (the `Integration (remediationorchestrator)` flake investigated above). RCA turned out to be more precise than my first pass in the earlier comment -- the race is between the RAR-approving test's naive `Get`-then-`Status().Update()` and `ApprovalCreator.persistApprovalRequest`'s own second `Status().Update()` (setting the initial `ApprovalPending`/`Decided`/`Expired` conditions right after RAR creation), not `RARReconciler`'s audit-condition write as I first guessed (that one only fires once `Status.Decision` is already non-empty, so it can't be racing the test's *first* write to Decision).

Fixed by adding a small `approveRAR` retry-on-conflict helper (mirroring the existing `RARReconciler.persistAuditRecordedCondition` / `injectTerminalStatus` convention) and using it at all 7 call sites that approve/reject a RAR across `lifecycle_test.go`, `override_flow_test.go`, `locking_integration_test.go`, and `events_test.go`. `go vet`, `golangci-lint`, and `gofmt` all clean; local envtest run got as far as suite bootstrap and then hit a pre-existing local-environment gap (this repo's shared postgres/redis/datastorage test infra needs Podman, which isn't configured on this box) unrelated to the change -- CI has that infra, so it'll be validated there.

Issue #1327, opened 2026-05-28
v1.6: Decouple LLM execution from KA orchestrator and consolidate session management into AF


jordigilh:
## Context

Today's architecture has three services in the investigation path:

```
User → AF (A2A/MCP, auth, session UI) → KA (orchestrator + LLM execution + K8s tools) → LLM provider
                                          ↑
                                    AA controller (autonomous)
```

KA currently bundles two distinct responsibilities:
1. **Orchestrator**: Session lifecycle, MCP tool registration, investigation state machine, audit emission, interactive/autonomous mode dispatch
2. **LLM Executor**: Prompt construction, LLM API calls, tool execution (kubectl_*, logs, prometheus), response parsing, alignment/grounding

These responsibilities have different trust and scaling profiles. The orchestrator is business logic that coordinates; the executor runs untrusted LLM-generated tool calls against the cluster.

## Proposal

Split KA into two components and redistribute:

### Component A: Orchestrator → Merge into AF
- Session lifecycle management (interactive, autonomous, observer modes)
- MCP tool registration and dispatch
- Investigation state machine (start, takeover, message, complete, cancel, status)
- Audit event emission
- Rate limiting, RBAC enforcement
- AA controller's `/api/v1/incident/analyze` endpoint

**Rationale**: AF already handles authentication, session management, and the A2A agent loop. The KA orchestrator is the natural continuation of that flow. Merging eliminates the AF→KA inter-service hop (TLS, SA tokens, circuit breakers, NetworkPolicies) for what is a 1:1 relationship.

### Component B: LLM Executor → Standalone sandboxed pod
- Prompt construction and LLM API calls
- Tool execution (kubectl_*, logs, prometheus queries)
- Response parsing, alignment, grounding
- Cluster-wide read RBAC (ClusterRole) stays with this pod
- Runs in isolated runtime with openshell sandbox, seccomp profile

**Rationale**: This is where untrusted LLM output drives tool execution. Isolating it in a sandbox with its own SA and restricted network egress reduces blast radius and satisfies FedRAMP SI-3/SI-7 (malicious code protection, software integrity).

### Communication: ACP (Agent Communication Protocol)
The merged AF+orchestrator communicates with the sandboxed executor via ACP. This is a natural fit because:
- v1.6 adds pluggable agent injection in investigation and workflow selection phases
- The executor becomes one of potentially many agents the orchestrator dispatches to
- ACP provides structured turn-based communication with streaming support

```
[...]

PR #2250, opened 2026-08-22
fix(kubernautagent): migrate AgentSession dispatcher to controller-runtime Reconciler (#2231)

### LINKED ISSUE #2231: KA Dispatcher.cancelOnDelete relies on a raw watch.Deleted event, same at-most-once delivery risk fixed in AF (#2214)
## Background

While root-causing the intermittent `IT-AF-2214-002` failure in PR #2222 (tracked/fixed by #2214), I reproduced the underlying issue on an isolated amd64 host loaded with the exact CI-built images from the failing run: AF's `AgentSessionTerminalCloseReconciler` used a raw `handler.Funcs.DeleteFunc` to react to `AgentSession` deletion, which is **best-effort, at-most-once delivery**. Under CPU contention (`--race --procs=4`), the single ephemeral delete event was dropped outright 6/23 times locally, with no recovery short of a client-go/apiserver watch-bounce relist (typically ~5-10 minutes).

The fix (merged in #2222) replaced that pattern with a standard Kubernetes finalizer (`agentsessionv1.TerminalCloseFinalizer`), so deletion is now observed via the same workqueue-backed, retried `Reconcile()` path already proven reliable for the Update case, eliminating the race by construction. Verified with 20/20 clean repro-loop runs after the fix (vs. 6/23 failures before).

## The same structural risk exists in KA, unaddressed

`internal/kubernautagent/agentsession/dispatcher.go`'s `Dispatcher.watchLoop` uses a **raw `watch.Interface`** (via `d.client.Watch(...)`) and reacts directly to `watch.Deleted` events in `handleEvent` -> `cancelOnDelete`, to stop the in-memory investigation goroutine when an `AgentSession` is deleted (e.g. AA's `DeleteForCascadeCancel`, or transitive owner-cascade deletion). This is the exact same "raw watch, at-most-once delivery" pattern that was just proven to drop events under load in AF's case.

## Why this wasn't fixed as part of #2214

Impact is bounded and self-correcting, unlike the AF case:
- If KA misses the delete event, the in-memory investigation goroutine simply keeps running until it reaches its own natural completion/timeout (burning LLM/tool budget for that duration), rather than permanently failing to converge to a required terminal state (which is what the AF bug did to `InvestigationSession`).
- KA's dispatcher doesn't have an obvious finalizer-based fix available: KA is not the resource's finalizer-holder candidate (AF/AA are the natural owners of that lifecycle contract), and adding a KA-side finalizer would introduce a new cross-service coupling that wasn't in scope for #2214.

## Ask

- Quantify the actual blast radius (how long can a missed cancel keep an investigation running -- is there an existing hard ceiling, e.g. `TimesOutAt`?).
- If the exposure is meaningful, consider a design fix (e.g. a lightweight periodic reconciliation/resync sweep in the dispatcher cross-checking live `AgentSession` state, or informer-based watch instead of a raw `watch.Interface`, which would get controller-runtime's retry/relist semantics for free).
- If bounded and acceptable, document the accepted risk explicitly next to `cancelOnDelete`'s existing doc comment, matching how bounded-degradation risks are documented elsewhere in this codebase (e.g. the IS TTL reaper backstop).

## References

- #2214 (the fix this finding came out of)
- PR #2222
- `docs/architecture/decisions/DD-AA-KA-001-agentsession-crd-http-removal.md`, "Post-merge correction" amendment (2026-08-21) -- has the full RCA and the "Cross-check (not fixed here, flagged for awareness)" note this issue tracks
- `internal/kubernautagent/agentsession/dispatcher.go`: `watchLoop`, `handleEvent`, `cancelOnDelete`

jordigilh:
## Summary

Closes #2231.

KA's `agentsession.Dispatcher` reacted to a raw `watch.Deleted` event via a hand-rolled `crclient.WithWatch` watch loop — delivered at-most-once, outside controller-runtime's workqueue retry machinery. A dropped delivery (confirmed live for AF's structurally-identical prior design, see DD-AA-KA-001's "Post-merge correction" amendment, PR #2222) left KA investigating a remediation nothing would ever read the result of, bounded only by `session.Store`'s 60-minute `maxSessionAge` backstop.

This PR migrates the Dispatcher to the standard controller-runtime `Manager` + `Reconciler` + finalizer pattern, mirroring AF's own proven `AgentSessionTerminalCloseReconciler` (`internal/controller/apifrontend/agentsession_close.go`) — full rationale in [DD-AA-KA-001's #2231 amendment](docs/architecture/decisions/DD-AA-KA-001-agentsession-crd-http-removal.md).

- `Dispatcher.Reconcile`/`reconcileDelete`/`SetupWithManager` replace the retired `Start`/`watchLoop`/`handleEvent`/ticker-driven `resync`. `tryDispatch`/`cancelOnTimeout` are still launched via goroutines so a real dispatch attempt never blocks controller-runtime's single reconcile worker.
- New `dispatchCleanupFinalizer` (`agentsessionv1.DispatchCleanupFinalizer`) defers actual `AgentSession` removal until `Reconcile` has observed the delete and stopped the in-memory investigation goroutine.
- `removeDispatchCleanupFinalizer` retries on `Conflict`: `cancelOnDelete`'s synchronous `ForceCancelByRemediationID` call can itself trigger a Status write via the `TerminalHook` on the same object before the finalizer removal runs, bumping `resourceVersion` out from under a stale copy — caught live by `UT-AA-2170-DELETE-001` during development.
- `cmd/kubernautagent/agentsession_wiring.go` now builds and starts a namespace-scoped `ctrl.Manager` for the dispatcher, mirroring `cmd/apifrontend/session_infra.go`.
- AA's `AgentSessionCreator.DeleteForRetry` strips the new finalizer (alongside `TerminalCloseFinalizer`) before deleting, preserving its synchronous-delete contract for the capacity-exceeded retry flow.
- RBAC: KA's ServiceAccount gains `update`/`patch` on `agentsessions` (metadata only) in both the Helm chart and the E2E hand-rolled manifest.

Also: posted a status-clarifying comment on #1327 (KA/AF-merge proposal), noting its ACP-executor-pod design is superseded by #1536's runtime-agnostic bring-your-own-OCI-agent model.

## Test plan

- [x] `docs/testing/2231/TEST_PLAN.md` — IEEE 829-2008 hybrid test plan
- [x] Unit: all 15 pre-existing `dispatcher_test.go` scenarios migrated to direct `Reconcile()` calls (mechanical, zero change in asserted outcome) + new `UT-AA-2231-001..005` for `Reconcile`/`reconcileDelete`/finalizer logic — `go test ./internal/kubernautagent/agentsession/... -race` (27/27 pass, 5x stability run)
- [x] Integration: new `IT-AA-2231-001/002` against a real envtest apiserver, proving `SetupWithManager` + a real `ctrl.Manager` (not just direct `Reconcile()` unit calls) dispatch and delete-reliability — `go test ./test/integration/kubernautagent/agentsessiondispatcher/... ` (2/2 pass, 3x stability run)
- [x] `go build ./...` clean
- [x] `golangci-lint run` clean (0 issues) on all touched packages
- [x] Full `internal/kubernautagent/...`, `pkg/aianalysis/...`, `api/agentsession/...` unit suites pass (no regressions)


jordigilh:
Filed jordigilh/kubernaut-operator#380 to track the corresponding RBAC parity update on the operator side: KA's `kubernautAgentInvestigatorClusterRole` in that repo currently has no `agentsessions`/`agentsessions/status` grant at all (a pre-existing gap, previously flagged but deferred in kubernaut-operator#368), and needs the `update`/`patch` verbs this PR adds to the Helm chart plus the base grant it was already missing.

PR #2409, opened 2026-09-16
fix(demo): remove E2E grounding alert from managed fleet monitoring

jordigilh:
## Summary
- remove the E2E-only `KAInteractiveFleetBridgeGrounding` rule from managed Prometheus demo manifests
- preserve the fixture in the legacy Prometheus path used by FP and Fleet E2E lanes
- update the manifest regression test

## Root Cause
The fleet demo uses the Thanos-enabled managed Prometheus helper, but Thanos was incorrectly used as the switch for an E2E-only `vector(1) > 0` grounding rule. This caused the alert to fire immediately on fresh hub and spoke clusters.

## Testing
- `go build ./...`
- `go test ./test/infrastructure`
- `git diff --check`


jordigilh:
Superseded by #2412, which combines this fleet fixture cleanup with the related AF fix.

PR #2411, opened 2026-09-16
fix(apifrontend): suppress unknown tool placeholder

### LINKED ISSUE #2410: AF: remove visible ellipsis fallback for unknown function calls
## Problem

The AF launcher emits a user-visible `...` reasoning message when a model function call is not present in `toolStatusMessages`.

In `pkg/apifrontend/launcher/part_converter.go`, `convertFunctionCall` currently returns:

```go
return a2a.TextPart{Text: "...\n\n"}
```

`emitPartViaBridge` then sends that text through `bridge.EmitReasoning`, so the Console can show a standalone ellipsis before the actual tool error, for example:

```text
...

Error: create RR for alert investigation: ...
```

## Provenance

This fallback was introduced by `0e75ab7f8` (`feat(af): replace verbose Processing... keepalive with compact dots`) as a replacement for the old visible `Processing...` text.

## Why it is obsolete

AF now has separate keepalive mechanisms:

- `StreamingExecutor.startKeepalive` emits metadata-only keepalive status events through `EmitKeepaliveDotSafe`.
- `EventBridge.EmitKeepaliveDot` intentionally leaves `Status.Message` nil, preventing chat-history/UI pollution.
- The status SSE path also has protocol-level `: heartbeat\n\n` comment frames.

The unknown-function fallback is therefore not needed for connection liveness and should not create user-facing reasoning content.

## Proposed change

Remove the visible fallback, preferably by returning no part for an unknown function call. Keep the metadata-only and protocol-level keepalive mechanisms unchanged.

Update `UT-AF-1189-106` to assert that unknown function calls do not emit a reasoning/status event.

## Acceptance criteria

- Unknown function calls produce no visible `...` reasoning text.
- Known tool-call status messages continue to render normally.
- Tool errors such as `Error: create RR for alert investigation: ...` remain visible.
- Metadata-only and protocol-level SSE keepalives remain unchanged.
- The relevant launcher tests pass.

jordigilh:
Closes #2410

## Summary

- Stop converting unknown model function calls into visible `...` reasoning text.
- Keep known tool-call status messages unchanged.
- Update the AF-1189 unit tests and test plan to assert suppression.

## Why

The visible ellipsis was introduced as a replacement for `Processing...` to keep the stream active. AF now has separate metadata-only keepalive events and protocol-level SSE heartbeat comment frames, so the fallback is unnecessary user-facing output.

## Validation

- `go test ./pkg/apifrontend/launcher` passed: 327 specs.
- `git diff --check` passed.

The metadata-only `EmitKeepaliveDot` path and `: heartbeat\n\n` protocol frames are intentionally unchanged.


jordigilh:
Superseded by #2412, which combines this AF fix with the related fleet fixture cleanup.

PR #2436, opened 2026-09-18
fix(fleet): avoid parsing MCP collision messages

jordigilh:
Stacked on #2435.

The shared MCP writer no longer classifies Create collisions by matching server-specific error text such as `field is immutable`. After a failed Create, it confirms the exact requested object through `resources_get`; only a successful matching read is translated to typed `AlreadyExists`, while unavailable or unsuccessful probes preserve the original error.

Validation:
- `make test-unit-shared-packages`
- `make build-workflowexecution`
- `make lint`


jordigilh:
Superseded: the hardening change now lands directly in PR #2435.

